PR #9, opened 2015-07-01
Target Java 7 for compilation

MichaelEvans:
If you have JDK 8 installed, you get a `com.android.dx.cf.iface.ParseException: bad class file magic (cafebabe) or version (0034.0000)` exception during a build. Setting the source compatibility version to 7 fixes that. 

I'm also not a Travis expert, but I believe this config change will run the builds on both JDK 7 and JDK 8 going forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ android:
     - extra-android-support
     - extra-google-m2repository
 
+jdk:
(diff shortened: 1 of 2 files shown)


felipecsl:
awesome. thanks!

Issue #7, opened 2015-07-01
Introduce a way to determine which path was used when multiple paths are used with @DeepLinks()


MichaelEvans:
Right now if you use something like `@DeepLinks({"example.com/deepLink/{id}", "example.com/anotherDeepLink"})`, you can determine which link was used by the presence of `id`. But if you did something like `@DeepLinks({"example.com/deepLink", "example.com/anotherDeepLink"})`, you'd have no mechanism to determine which path was used to get here.

felipecsl:
it looks like one possible solution would be to expose in the intent extas the original URI string used for matching, eg.: `"example.com/deepLink/{id}"` or `"example.com/anotherDeepLink"` in this case.

PR #10, opened 2015-07-01
Expose URI that was deeplinked through a parameter

MichaelEvans:


MichaelEvans:
Addresses #7

felipecsl:
I'd like to see a test around this change to guarantee that the generated intent has this new key in the bundle. Please let me know if you need directions on how to do it

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -187,6 +187,7 @@ private void generateDeepLinkActivity() throws IOException {
         .endControlFlow()
         .addStatement("parameterMap.put(queryParameter, uri.getQueryParameter(queryParameter))")
         .endControlFlow()
+        .addStatement("parameterMap.put(DeepLink.URI, hostPath)")
```
cdeonier: We should have an intact URI with the scheme, so we should replace `hostPath` with `uri.toString()`.

MichaelEvans:
@felipecsl Yeah, if you could give me some pointers on how I could test this, I'd gladly write some or it. I wanted to submit one, but I wasn't sure how best to test that.

felipecsl:
@MichaelEvans looks like none of the existing test are suitable to cover this test scenario. I'll set up some Robolectric tests that will make this easier. Hang tight!

MichaelEvans:
@felipecsl added some assertions to the tests from #11 to verify the behavior

cdeonier:
:boat:

felipecsl:
yay! thanks @MichaelEvans

MichaelEvans:
woot! thanks!

PR #11, opened 2015-07-03
Adds intent tests for MainActivity

felipecsl:
@cdeonier 

This should help @MichaelEvans write some tests for #10


cdeonier:
:+1:

Issue #14, opened 2015-07-06
Parameters can't have '_' and '-' symbols, but should


Javernaut:
I was about to use your library in my project, but it didn't recognize my url. As was found out, the regexp for parameters is too restrictive, as it forbids _ and - symbols.

My target url was like this:
someschema://somepath/an-identifier-with-dashes

Please, add this functionality.

MichaelEvans:
@felipecsl @cdeonier is there any reason that the param value regex is so restrictive? I would like to use hyphenated slugs in my paths as well. (I can put up a PR with a test if there's no real reason)

cdeonier:
@Javernaut @MichaelEvans Just an oversight. We should allow those symbols in the parameter. I'll go ahead and add this in.

cdeonier:
Oh, you've got it up already, great. I'll review that.

PR #15, opened 2015-07-06
Add support for hyphens and underscores as values for deeplink params

MichaelEvans:
Addresses #14 (Assuming there's no real reason why these characters were excluded)


cdeonier:
:+1:

felipecsl:
:+1: :+1:

MichaelEvans:
thanks again @cdeonier and @felipecsl!

PR #19, opened 2015-07-07
Adds support for multiple schemes

### LINKED ISSUE #6: Support multiple schemes
Modify `@DeepLink` and `@DeepLinks` to also require the `scheme://` in the URI definition, so we can suport more than one scheme in `DeepLinkActivity`.

felipecsl:
Fixes #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Sometimes you'll have an activity that should handle several kinds of deep links
 `@DeepLinks` annotation to register multiple deep links on an activity:
 
 ```java
-@DeepLinks({"example.com/deepLink/{id}", "example.com/anotherDeepLink"})
+@DeepLinks({"foo://example.com/deepLink/{id}", "foo://example.com/anotherDeepLink"})
 public class MainActivity extends Activity {
   @Override protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
@@ -61,7 +61,7 @@ You can also annotate static methods that return an `Intent`. `DeepLinkDispatch`
 method to create that `Intent` and use it when starting your activity via that registered deep link:
 
 ```java
-@DeepLink("example.com/methodDeepLink/{param1}")
+@DeepLink("foo://example.com/methodDeepLink/{param1}")
 public static Intent intentForDeepLinkMethod(Context context) {
   return new Intent(context, MainActivity.class).setAction(ACTION_DEEP_LINK_METHOD);
 }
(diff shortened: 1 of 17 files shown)


felipecsl:
@cdeonier

README.md
```diff
@@ -38,7 +38,7 @@ Sometimes you'll have an activity that should handle several kinds of deep links
 `@DeepLinks` annotation to register multiple deep links on an activity:
 
 ```java
-@DeepLinks({"example.com/deepLink/{id}", "example.com/anotherDeepLink"})
+@DeepLinks({"foo://example.com/deepLink/{id}", "foo://example.com/anotherDeepLink"})
```
cdeonier: Maybe include an example where the scheme is optional (ie, omitted).

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.java
```diff
@@ -27,76 +31,75 @@
...
-    this.regex = generateLookupString(hostPath);
+  DeepLinkEntry(String uri, Type type, Class<?> activityClass, String method) {
     this.type = type;
     this.activityClass = activityClass;
     this.method = method;
+    this.parameters = parsePathParameters(uri);
+    this.regex = schemeHostAndPath(uri).replaceAll(PARAM_REGEX, PARAM_VALUE);
   }
 
-  public String getRegex() {
```
cdeonier: Not sure why we got rid of the getter/setters here. I understand that it's still accessible as a member variable, but my preference is that it stays accessed through these methods.

README.md
felipecsl: My plan was to make the scheme mandatory (even though this is a breaking change) in order to make it a valid URI

felipecsl: ... and also remove ambiguity

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.java
felipecsl: ok! I'll bring back the getters!

cdeonier:
:+1:

README.md
MichaelEvans: If this is a breaking change, is it worth throwing an exception for a missing scheme in the processor?

felipecsl: Compilation will fail if the URI is invalid. It is now being checked by `DeepLinkAnnotatedElement`

MichaelEvans: Ok cool! (Was reading this on mobile and missed that)

PR #20, opened 2015-07-07
Set up automatic snapshot deployment

felipecsl:
This should allow Travis to automatically deploy snapshot releases to Sonatype after every change on master. Build script shamelessly copied from Retrofit
The encrypted env vars in the travis yml are the deployment username and password.

@cdeonier

### DIFF
--- a/.buildscript/deploy_snapshot.sh
+++ b/.buildscript/deploy_snapshot.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# Deploy a jar, source jar, and javadoc jar to Sonatype's snapshot repo.
+#
+# Adapted from https://coderwall.com/p/9b_lfq and
+# http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/
+
+SLUG="square/retrofit"
+JDK="oraclejdk8"
+BRANCH="master"
+
+set -e
+
+if [ "$TRAVIS_REPO_SLUG" != "$SLUG" ]; then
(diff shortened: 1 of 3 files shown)


.buildscript/deploy_snapshot.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#
+# Deploy a jar, source jar, and javadoc jar to Sonatype's snapshot repo.
+#
+# Adapted from https://coderwall.com/p/9b_lfq and
+# http://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/
+
+SLUG="square/retrofit"
```
cdeonier: do we need to change this?

gradle.properties
```diff
@@ -1,4 +1,4 @@
-PROJECT_VERSION=1.1.0
+PROJECT_VERSION=1.2.0-SNAPSHOT
```
cdeonier: So when people include dependencies, it will always be suffixed with "SNAPSHOT"?

.buildscript/deploy_snapshot.sh
felipecsl: oops :)

gradle.properties
felipecsl: yes, if they want to get the pre-release version. latest release is 1.1.0 but current master is 1.2.0-SNAPSHOT (pre-release) and next release will be 1.2.0. Once we release 1.2.0, this will be 1.3.0-SNAPSHOT (or whatever the next version is gonna be), and so on

cdeonier: :+1:

PR #21, opened 2015-07-07
Downgrade Okio

felipecsl:
Latest release of OkHttp depends on Okio 1.4.0, so let's not use a
version that's newer than that to prevent issues.

@cdeonier


cdeonier:
:+1:

PR #29, opened 2015-07-14
Adds both annotations to supported annotation types

### LINKED ISSUE #24: DeepLinkActivity not generated
I'm having some issues implementing this in my build. I've one successful build where DeepLinkActivity was generated, but most of the time it is not getting generated. 

I have a few other apt libraries in my build so I wonder if that is causing conflicts?

Is there a way I can debug why DeepLinkActivity is not getting generated?

felipecsl:
@cdeonier 

Fixes #24


cdeonier:
:+1:

PR #30, opened 2015-07-17
Apply a stricter regex when scanning DeepLinkRegistry

MichaelEvans:
Right now, the regex scheme set up in DeepLinkEntry allows for DeepLinkRegistry.parseUri to return early if there's a subset scheme already registered.

IE: 
- IndexActivity -> "foo://items"
- ShowActivity -> "foo://items/123"

Appending the `$` at the end of the regex should ensure that the whole String matches, rather than just the beginning.


felipecsl:
thank you!

PR #35, opened 2015-07-27
More relaxed path matching regex

### LINKED ISSUE #34: Urls with spaces aren't matched
with a url that contains spaces for example that are replaced with "%20"  doesn't get matched : 
for example with a url `http://www.example.com/action/search%20paris%20France%2092130"` and with an annotation on an Activity like : `@DeepLinks({"http://www.example.com/action/{searchAddress}"})`
no match is found and `DeepLinkEntry entry = registry.parseUri(uriString);` in DeepLinkActivity always returns a null entry and so 
  ```
   else {
      notifyListener(true, uri, "No registered entity to handle deep link: " + uri.toString());
      finish();
    }
   is called instead. 
is this not supported ? 

Thanks.

felipecsl:
Our path matching regex was too strict. Allow a few more characters.
List of URL code points: https://url.spec.whatwg.org/#url-code-points

Fixes #34 

@cdeonier

### DIFF
--- a/deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.java
+++ b/deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.java
@@ -27,7 +27,7 @@ import java.util.regex.Pattern;
 
 final class DeepLinkEntry {
 
-  private static final String PARAM_VALUE = "([a-zA-Z0-9_-]*)";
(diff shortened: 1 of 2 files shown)


SeanA208:
For reference https://www.debuggex.com/r/j7NoGIn8C7yGEVfB

felipecsl:
@cdeonier updated!

cdeonier:
:+1:

PR #45, opened 2015-08-25
Fixes NPE for invalid Uri

felipecsl:


felipecsl:
@cdeonier

cdeonier:
:+1:

elihart:
lgtm

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.java
```diff
@@ -104,7 +104,7 @@ private static String parsePath(DeepLinkUri parsedUri) {
 
   boolean matches(String inputUri) {
     DeepLinkUri deepLinkUri = DeepLinkUri.parse(inputUri);
```
elihart: so it's completely valid that the `parse` method returns null? should it be annotated with nullable? As far as the source of that crash, should we care about why the input uri is not valid?

felipecsl: Yeah, from the javadocs:
```
  /**
   * Returns a new {@code OkUrl} representing {@code url} if it is a well-formed HTTP or HTTPS URL,
   * or null if it isn't.
   */
```

That's just a copy of OkHttp's OkHttp's [HttpUrl](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java) with minor changes to allow any scheme, so I didn't want to change it too much

PR #46, opened 2015-08-25
Removes the need to use different annotations

felipecsl:
We can achieve the same functionality by using a single annotation `@DeepLink`, where you can use both single uris (eg.: `@DeepLink("http://example.com/foo")`) or multiple (eg.: `@DeepLink({"http://example.com/foo", http://example.com/bar"})`)

@cdeonier


cdeonier:
Assuming you checked the singular case, :+1:

felipecsl:
yep! thanks!

PR #47, opened 2015-08-25
Update readme and sample to use latest android-apt version

MichaelEvans:


MichaelEvans:
Also, seems like there's some random spaces in some of these files?

felipecsl:
WAT?! also apparently doenst work with apt 1.7 as it failed to generate the classes

MichaelEvans:
@felipecsl yeah I see that...not sure why, `./gradlew check assemble` worked for me locally. Ideas?

felipecsl:
apparently if you assemble first and check after, it works fine. I'll merge this and update the travis script manually. thanks!

MichaelEvans:
:+1: thanks! Didn't even notice that. Good eye!

PR #50, opened 2015-09-21
block non-static method to avoid runtime exception

imlunacat:
If a non-static is annotated with `@DeepLink` and invoked by deeplink. Application would crash in method invocation.
i.e.
```java
// cause crash if airbnb://host/20 is retrieved
@DeepLink("airbnb://host/{arbitraryNumber}")
public Intent intentFromNoStatic(Context context) {
  return new Intent(context, MainActivity.class).setAction(ACTION_DEEP_LINK_COMPLEX);
}
```

Throw Exception in compile-time rather than crash in runtime.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -76,6 +76,14 @@ public class DeepLinkProcessor extends AbstractProcessor {
             DeepLink.class.getSimpleName());
       }
 
+      if (kind == ElementKind.METHOD) {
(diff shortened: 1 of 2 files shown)


felipecsl:
Looks like checkstyle failed, can you fix it please? https://travis-ci.org/airbnb/DeepLinkDispatch/jobs/81404882
Also please add a unit test!
Thanks!

imlunacat:
Hi felipecsl,
  It seems checkstyle fail is caused by `DeepLinkUri.java`. These fail should be fixed in this pull-request ?

felipecsl:
you're right. that has been fixed on master already. just needs a test now!

imlunacat:
Hello,
  Is there any update about this PR ?
  It seems all checks are passed and unit test is added now ;)

felipecsl:
Looks good now. Mergin. Thanks!

PR #57, opened 2015-10-24
Fixed scheme setter in DeepLinkUri Builder.

arpitanand:
The changes done to HttpUrl class missed out setting the scheme in Builder.


felipecsl:
Thanks!

PR #59, opened 2015-11-02
Get a query parameter that contains the brackets

mapyo:
Thank you for this great library!

I could not get a query parameter that contains the brackets.

For example,
```
airbnb://foo/baz?marsh[max]=mallow
```

`marsh[max]` is null.

The reason for this, please refer to the following.

https://github.com/airbnb/DeepLinkDispatch/blob/759cd16/deeplinkdispatch-processor/src/test/resources/DeepLinkActivity.java#L41

sample
```
import android.net.Uri;

String urlString = "airbnb://foo/baz?marsh[max]=mallow";
Uri uri = Uri.parse(urlString);

uri.getQueryParameter("marsh[max]");  // null

uri.getQueryParameterNames(); // can get key(marsh[max])
```

`android.net.Uri` package can not get a query parameter that contains the brackets.

For this reason, I have sent a pull request.

Thank you.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -198,15 +198,19 @@ public class DeepLinkProcessor extends AbstractProcessor {
         .addStatement("$T uri = getIntent().getData()", ClassName.get("android.net", "Uri"))
         .addStatement("String uriString = uri.toString()")
         .addStatement("DeepLinkEntry entry = registry.parseUri(uriString)")
+        .addStatement("DeepLinkUri deepLinkUri = DeepLinkUri.parse(uriString)")
         .beginControlFlow("if (entry != null)")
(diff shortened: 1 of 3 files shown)


felipecsl:
I think we can probably fix this by using `DeepLinkUri` to parse the intent data query parameters instead of `android.net.Uri`. That solution would be preferable since it's simpler. Please feel free to give it a shot and update the PR

mapyo:
@felipecsl 
Thank you for your comment!
I updated this PR.

Please confirm :octocat:

felipecsl:
Nice! Looking great, thanks @mapyo 
Last thing, please rebase the commits then I'll go ahead and merge it.

mapyo:
@felipecsl 
I rebased.
Thank you!

felipecsl:
Perfect. Thanks a lot!

mapyo:
:smile:

mapyo:
@felipecsl 
When is it scheduled to be released?

felipecsl:
@mapyo we don't have a specific date yet, but I believe 1.5.0 should be out sometime later this month.
In the meantime you can use the snapshot version `1.5.0-SNAPSHOT`

mapyo:
Thank you!

PR #60, opened 2015-11-09
Simplify DeepLinkLoader + some cleanup

felipecsl:
Turns out that the `Loader` interface is not needed at all, since we can reference `DeepLinkLoader` directly from `DeepLinkActivity`. Also removed `DeepLinkRegistry` and moved a method into the loader class. 
Updated some dependencies and some minor cleanup around.

@cdeonier


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -116,25 +119,41 @@ private void error(Element e, String msg, Object... args) {
     messager.printMessage(Diagnostic.Kind.ERROR, String.format(msg, args), e);
   }
 
-  private void generateRegistry(List<DeepLinkAnnotatedElement> elements) throws IOException {
+  private void generateDeepLinkLoader(List<DeepLinkAnnotatedElement> elements) throws IOException {
     MethodSpec.Builder loadMethod = MethodSpec.methodBuilder("load")
```
cdeonier: Move this down right above the `for` loop on 132 to keep relevant localized by each other.

sample/src/main/AndroidManifest.xml
```diff
@@ -26,8 +26,8 @@
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="airbnb" />
-                <data android:scheme="http" />
+                <data android:scheme="dld" />
+                <data android:scheme="http" android:host="example.com" />
```
felipecsl: renaming/updating this to avoid people from unadvertedly copy/pasting the sample code into their apps, which would cause weird problems by using the airbnb scheme and a catchall http intent filter

cdeonier: :+1:

cdeonier:
great simplification, :+1:

Issue #61, opened 2015-11-11
Parameters regex should not match strings containing "/"


RomainPiel:
In my opinion this test should pass:

```java
@Test public void multiplePathParams() {
    DeepLinkEntry entry = deepLinkEntry("airbnb://foos/{foo}");

    Map<String, String> parameters = entry.getParameters("airbnb://foos/baz/qux");
    assertThat(parameters)
      .isEmpty();
  }
```

but it's not. It finds the parameter `foo` with the value `baz/qux`.
The funny thing is that I can't repro with `airbnb://{foo}` and `"airbnb://baz/qux"`.

felipecsl:
Fixed via #62

PR #62, opened 2015-11-11
a parameter should not contain a forward slash, corrected param value regex

RomainPiel:
See #61 for more detail about the issue.

The problem was is the regex `%-~` means "any character from % to ~". What we want is an exact match to `-` so `\\-` works better.


felipecsl:
Oh good catch! Thanks!

Issue #76, opened 2016-01-15
Allow a result to be returned to the caller


sheaam30:
If `DeepLinkActivity` receives an intent that was started with `startActivityForResult` there is currently no way to return a result to the Activity that sent the intent.

felipecsl:
Sounds good to me. Please feel free to send a PR

andrewlord1990:
It looks like this issue can be closed.

PR #77, opened 2016-01-15
as-add-result-support

sheaam30:
- Add support for if the Activity was called with startActivityForResult then we want to forward results back to the caller.  #76


felipecsl:
Looks good to me. Just fix the failing test and I'll merge

sheaam30:
Addressed comments @felipecsl  :)

felipecsl:
Thanks!

PR #78, opened 2016-01-27
Update readme on static method deeplinking

geobio:
Update readme to explicitly declare that a static method must only receive one Context argument.

Originally I tried to use a method with more than one argument, and took a lot of debugging through error messages, to realize that it's because it needs exactly one Context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ public class MainActivity extends Activity {
 
 ### Method Annotations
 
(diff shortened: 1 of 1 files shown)


felipecsl:
Good point. We should also update the `DeepLinkProcessor` to perform that validation during compile time so it doesn't break at runtime. Can you also fix that?

PR #82, opened 2016-03-10
Allowing e-mails in the regular expression validation

Dominiquini:


felipecsl:
@Dominiquini please add a test that checks for the use case you want to handle here

felipecsl:
Closed via #99

PR #84, opened 2016-03-15
Do not fail on packages with uppercase

### LINKED ISSUE #83: Deeplink loader class is incorrectly generated when project uses capital letters in Java package names
In the app I work on, we use capital letters in the package names, and DeeplinkDispatch incorrectly thinks that the package is a class, and that the actual activity is a nested class.

So, it generates DeepLinkLoader with "import com.example.Package"

and in the load() method:
new DeepLinkEntry("deeplink://path1", DeepLinkEntry.Type.CLASS, Package.Activity.class, null)

which fails to compile. 

If I change that package name to small caps, then I can successfully compile the app, and the deeplinks work as well. 

Could you please change the code to not automatically consider packages with capital letters as classes?

I believe the relevant code is in: ClassName.bestGuess() method, which is actually from squareup/javapoet

felipecsl:
Fixes #83 

@cdeonier


cdeonier:
:+1:

PR #87, opened 2016-03-22
DeepLinkDelegate

### LINKED ISSUE #86: Proposal: Make generated DeepLinkActivity optional
## The problem

Currently, all deeplink handling is done in a generated `DeepLinkActivity` that you must add to your manifest. While this is a convenient mechanism for easy dropping in, it's not very extensible if you have any requirements around your deeplinking. In particular, you ultimately don't control the entry point to your app, and can't run any sort of preconditions. 

## Example

Any app that has required login to proceed faces this issue. You receive a deeplink, check if the user is logged in, if yes then proceed, otherwise take them to the login screen. Unfortunately this doesn't allow for that kind of handling since there's no control over the generated activity. The dispatch gets done, regardless of the app state. 

## Proposed solutions

The generated `DeepLinkActivity` is actually well-suited to extraction to a class. My proposal is this: generate a `Dispatcher` class like the following:

```java
public final class Dispatcher {
    public static DispatchResult dispatchFrom(Activity activity) {
        // Check activity != null...
        // Do all the dispatch handling...

        // Return the result
        return new DispatchResult(isError, uri, errorMessage);
    }
}
```

Where `DispatchResult` is a simple data class representing the information you'd otherwise send to the `DeepLinkReceiver`.

This would allow you to take advantage of the main benefit of the library (generated uri handling) while allowing you to keep control over entry points.

```java
public class LaunchActivity extends Activity {
    @Override
    public void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        
        if (isDeepLink() && requiredConditionsAreValid() && isLoggedIn()) {
            DispatchResult result = Dispatcher.dispatchFrom(this);

            if (!result.isSuccessful()) {
                // Do something about it
            else {
                // Done? Maybe? ¯\_(ツ)_/¯
                return;
            }
        }
    }
}
```

You could keep the currently existing implementation as well for convenience and just have it call through  to the `Dispatcher`. This would also make it easy to test, especially if you opted for a new instance vs. static methods (which opens up easy injection and handling via Dagger/other DI implementations).

The alternative idea would be to add some form of precondition callback that the developer could register. This isn't as ideal though since it's fairly limited.

Names are totally up for discussion by the way! Curious to hear people's thoughts on this. If it's something you'd be for, can gladly contribute a PR for this.

hzsweers:
This implements what was discussed in #86 by using a generate `DeepLinkDelegate` to handle dispatches instead of maintaining the logic solely in a generated `DeepLinkActivity`.

Three parts:
- Generate DeepLinkDelegate class with its single static `dispatchFrom(Activity)` method.
- Slim down the generated DeepLinkActivity to just use this delegate
- Generate a DeepLinkResult class to use on the fly to return results from `dispatchFrom()`
  - This is a little wonky because the current API is just a Java library and not an AAR. Generating it is the only current way to make this class available in generated code.

Now you can simply call this from your own activity to handle a dispatch:

```java
DeepLinkResult result = DeepLinkDelegate.dispatchFrom(this);

if (!result.isSuccessful()) {
  // Do whatever
  // result.uri();
  // result.error();
} else {
  // Yay. The activity started, probably don't actually want to do anything in this case
}
```

Open questions:
- API/naming
- Should `DeepLinkDelegate` be an instance or just have its one static method?
  - Is there any benefit to having an instance? Testing, helpful with DI?
- Should `DeepLinkActivity` generation be disable-able? If so, how?
  - Even though not adding it to the manifest closes it off, some might still not like having it at all.
  - Alternative is to remove it entirely.
- Where does `DeepLinkLoader` fit in? Should it remain separate?

Resolves #86


hzsweers:
My personal thoughts on the open questions:

- I like this naming.
- API seems a little weird in that it can return a result even if it was successful, so the calling activity could do something after the new one was launched. I suppose there's nothing wrong with that if the user is sensible about it.
- I prefer static right now, and maybe considering other options later if the need arises for it.
- Would be nice to have it disable-able, but don't know how that could be controlled dynamically. Would personally be fine with removing it entirely, though that would be a breaking API change.
- Fine with the loader being separate for testing and separation of concerns, but it does seem like something that could be as easily pulled into the delegate class or even inlined if we wanted.

felipecsl:
looking pretty good so far. just need to make checkstyle happy.
@cdeonier might want to take a look as well

hzsweers:
Finally fixed all the checkstyle issues and CI's happy

hzsweers:
New thought: should `DeepLinkResult#errorMessage()` maybe just be shortened to `error()`?

felipecsl:
`errorMessage` -> `error` :+1: 
* Sounds like it could be a good simplification for the `DeepLinkDelegate` to also notify the broadcast receiver.
* Re. taking a `DispatchResultCallback` I'm not that sure. Maybe for a follow-up PR

hzsweers:
Restructured more, sorry it ended up being a decent amount of change

- `error()` instead of `errorMessage()`
- dispatch now returns a result and also handles notifying listeners
- refactor constants and create `DeepLinkActivity` annotation
  - This holds those constants in a non-generated place and also makes it easy to indicate to the processor that it doesn't need to generate an activity when it's used. This prevents extra entry points or unnecessarily generated code.

It's a breaking change, but I think one worth making anyway.

felipecsl:
Alright this is looking pretty good to me. Do you mind squashing the commits before we merge?

hzsweers:
Done

felipecsl:
Thanks! I'll bump the pre-release to `2.0.0-SNAPSHOT`

hzsweers:
awesome, thanks!

PR #89, opened 2016-04-01
Renames

hzsweers:
#### Rename DeepLinkActivity -> DeepLinkHandler
  - More generic naming so it doesn't look weird to have it on non-activities (fragments, controllers, wherever)

#### Rename DeepLinkDispatchActivity -> DeepLinkActivity
  - Preserves compatibility for anyone that used it before and means they don't have to update names if they want to keep it.


felipecsl:
:shipit:

hzsweers:
🚢

PR #94, opened 2016-05-18
Fixes regression in intent data and action

felipecsl:
Introduced by #87 

@cdeonier @hzsweers


hzsweers:
Good catch, my bad :). Lgtm

PR #102, opened 2016-06-06
Require a non-zero length param value

MichaelEvans:
Right now (and I believe as of 1.5.0), we're seeing a bug where `DeepLinkEntry` matches against Uris with empty params improperly.

For instance - If you have:

```
dld://foo
dld://foo/{id}
```

You'll find that the second URI is matched - with an empty id, rather than matching the first path. Additionally, you'll see that routes like `dld://foo/{id}/bar` are incorrectly matching against things like `dld://foo//bar`. Changing the Regex operator to require a character match would fix these two issues.

Also, for an added bonus - I think the first case is obscured based on the order that the entries are listed in the Loader - which is based on the name of the Activity.


hzsweers:
:+1:

felipecsl:
Thank you!

Issue #104, opened 2016-06-07
LocalBroadcast Creates an NPE when sourceIntent's data Uri is null.


MinceMan:
In the following code you can see that a null uri will be passed down to notifyListener and `uri.toString()` will be called on it.

```java
public final class DeepLinkDelegate {

  // code

  public static DeepLinkResult dispatchFrom(Activity activity) {
    if (activity == null) {
      throw new NullPointerException("activity == null");
    }
    Intent sourceIntent = activity.getIntent();
    Uri uri = sourceIntent.getData();
    if (uri == null) {
      return createResultAndNotify(activity, false, null, "No Uri in given activity's intent.");
    }

    // Code
  }

  private static DeepLinkResult createResultAndNotify(Context context, final boolean successful, final Uri uri, final String error) {
    DeepLinkResult result = new DeepLinkResult(successful, uri, error);
    notifyListener(context, !successful, uri, error);
    return result;
  }

  private static void notifyListener(Context context, boolean isError, Uri uri, String errorMessage) {
    Intent intent = new Intent();
    intent.setAction(DeepLinkHandler.ACTION);
    intent.putExtra(DeepLinkHandler.EXTRA_URI, uri.toString());  // <-- NPE here
    intent.putExtra(DeepLinkHandler.EXTRA_SUCCESSFUL, !isError);
    if (isError) {
      intent.putExtra(DeepLinkHandler.EXTRA_ERROR_MESSAGE, errorMessage);
    }
    LocalBroadcastManager.getInstance(context).sendBroadcast(intent);
  }
}
```

MinceMan:
Would the following be the fix? Or do we want to broadcast that error no matter what?

'''java
  private static DeepLinkResult createResultAndNotify(Context context, final boolean successful, final Uri uri, final String error) {
    DeepLinkResult result = new DeepLinkResult(successful, uri, error);
    if (uri != null) notifyListener(context, !successful, uri, error);
    return result;
  }
```

MinceMan:
I'll open a PR once I get a response on the correct fix :)

felipecsl:
I believe that in the case of a launched deep link, theoretically the Intent data cannot be `null`, but I agree this is wrong. We should check if the `uri` is not `null` before calling `toString()` on it. Can you send a PR?

MinceMan:
Will do.

PR #105, opened 2016-06-08
Fix for #104

MinceMan:
Fix for a NullPointerExeception when sending through an activity with no data.

When using DeeplinkDelegate on an activity that is launched through the launcher it will create a NPE as the activity doesn't have data.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -280,7 +280,7 @@ public class DeepLinkProcessor extends AbstractProcessor {
         .addParameter(String.class, "errorMessage")
         .addStatement("$T intent = new Intent()", ANDROID_INTENT)
         .addStatement("intent.setAction($T.ACTION)", DeepLinkHandler.class)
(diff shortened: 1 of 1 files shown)


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -280,7 +280,7 @@ private void generateDeepLinkDelegate() throws IOException {
         .addParameter(String.class, "errorMessage")
         .addStatement("$T intent = new Intent()", ANDROID_INTENT)
         .addStatement("intent.setAction($T.ACTION)", DeepLinkHandler.class)
-        .addStatement("intent.putExtra($T.EXTRA_URI, uri.toString())", DeepLinkHandler.class)
+        .addStatement("intent.putExtra($T.EXTRA_URI, uri != null ? uri.toString() : $S)", DeepLinkHandler.class, "")
```
felipecsl: why not use "" directly?

MinceMan: I thought it looked less clean escaping the quotes. I can change it if you want.

felipecsl: no worries, this is fine. just curious ;)

felipecsl:
thanks!

PR #106, opened 2016-06-08
Adding singleTop support

MinceMan:
A singleTop activity doesn't get deeplink info through `getIntent()` on the activity. Instead it gets it from a call on `onNewIntent(Intent intent)` and that intent must be handed to DeeplinkDelegate. Here I overloaded the `dispatchFrom` method to support handing through both the activity and the intent that should be used.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -320,7 +320,20 @@ public class DeepLinkProcessor extends AbstractProcessor {
         .beginControlFlow("if (activity == null)")
         .addStatement("throw new $T($S)", NullPointerException.class, "activity == null")
         .endControlFlow()
-        .addStatement("$T sourceIntent = activity.getIntent()", ANDROID_INTENT)
+        .addStatement("return dispatchFrom(activity, activity.getIntent())")
+        .build();
+
+    MethodSpec dispatchFromMethodWithIntent = MethodSpec.methodBuilder("dispatchFrom")
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .returns(DEEPLINKRESULT)
+        .addParameter(ClassName.get("android.app", "Activity"), "activity")
+        .addParameter(ClassName.get("android.content", "Intent"), "sourceIntent")
+        .beginControlFlow("if (activity == null)")
+        .addStatement("throw new $T($S)", NullPointerException.class, "activity == null")
+        .endControlFlow()
+        .beginControlFlow("if (sourceIntent == null)")
+        .addStatement("throw new $T($S)", NullPointerException.class, "sourceIntent == null")
+        .endControlFlow()
         .addStatement("$T uri = sourceIntent.getData()", ANDROID_URI)
         .beginControlFlow("if (uri == null)")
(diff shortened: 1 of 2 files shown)


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -319,7 +319,20 @@ private void generateDeepLinkDelegate() throws IOException {
...
+        .addStatement("return dispatchFrom(activity, activity.getIntent())")
+        .build();
+
+    MethodSpec dispatchFromMethodWithIntent = MethodSpec.methodBuilder("dispatchFrom")
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .returns(DEEPLINKRESULT)
+        .addParameter(ClassName.get("android.app", "Activity"), "activity")
+        .addParameter(ClassName.get("android.content", "Intent"), "sourceIntent")
+        .beginControlFlow("if (activity == null)")
+        .addStatement("throw new $T($S)", NullPointerException.class, "activity == null")
```
felipecsl: we're already checking for null activity on line 319

felipecsl:
how is this adding support for singleTop? I don't see where that is happening

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
MinceMan: I'll remove the check from 319 so we just use this one.

MinceMan:
For singleTop, singleInstance and singleTask the deeplink intent comes through `Activity.onNewIntent(Intent intent)` and is not available from `Activity.getIntent()`. So using the `DeepLinkDelegate` with the new change looks like this.

```java
    @Override
    protected void onNewIntent(Intent intent) {
        DeepLinkDelegate.dispatchFrom(this, intent);
    }
```

I didn't add support for the `DeepLinkActivity` as I didn't think that would be a legitimate use case.

felipecsl:
Got it. Can you also update the `DeepLinkDelegate.java` file in `src/test/resources` so it correctly reflects the new method?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
MinceMan: I just realized that if I remove 319 then 322 could null out. So while it is a redundant check it is needed for good error reporting. Can still remove it if you want but I don't think we do.

felipecsl:
Thanks!

OleksandrKucherenko:
One note: `onNewIntent` received before the `onResume`, so if you want to recover UI state based on deeplink request, than you need to store in some variable and process in `onPostResume()`.

why in `onPostResume` - keep in mind that activity may have multiple fragments, and fragments recover own state and receive `onResume` when called activity `super.onPostResume()`... 
take a look on full lifecycle of activity and fragments - https://plus.google.com/u/0/+OleksandrKucherenko/posts/WtmnBRohzs1

JulienRaad:
Hello
How is this achieved ? I am still getting a new activity every time even though the activity is already opened

JulienRaad:
singleTop did not work, singleTask works well

Issue #93, opened 2016-05-10
Custom path matching regex


fernandospr:
Currently I have seen that the regex (DeepLinkEntry.PARAM_VALUE) is "([a-zA-Z0-9_#!+%~,\\-\\.\\$]*)".
However I have deeplinks that contains other characters, for example '&'.
So they will not match.

The proposal is to allow the client to customize this regex.

Thanks.

felipecsl:
I don't think we wanna make the regex customizable at runtime, but please feel free to submit a pull request that allows `&` on the path regex

fernandospr:
Do you see any problems making this externally configured? Like using a meta tag in the manifest? I can make a pull request to include '&', however tomorrow maybe I will need more, I don't know maybe ¡ ¿ @ etc.

felipecsl:
I believe this regex should be following the URI specification for the [path component](https://tools.ietf.org/html/rfc3986#page-22). While the RFC does not explicitly state which characters are allowed, the rules are not supposed to change.
The problem here is that our regex is too restrictive and should just be fixed to allow more special characters

PR #111, opened 2016-06-28
Customize the path matching regex, add &?()

padfoot27:
Customize the PARAM_VALUE to accommodate for URLs that contains &,?,(,)


MichaelEvans:
I don't know about parenthesis, but I don't think you want to include ? or & since they are used to delimit query parameters.

padfoot27:
You can check out the comments in this issue - https://github.com/airbnb/DeepLinkDispatch/issues/93.

PR #126, opened 2016-10-05
Allow for DeepLinkDispatch to be used in multiple modules

### LINKED ISSUE #90: Cannot apply into multi-modules project
The files generated by apt have the file names that are hard coded like following.  

image

image

As a result, the library can only apply into the project owning only one module app. Otherwise, if applied into one main app and other depend library module, there would be more than one generated files with same names. And that would lead to dex error.

Would you please upgrade the current implementation to support multi-modules project?

felipecsl:
Fixes #90 by scoping the generated classes to the same package of the
annotated classes.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.java
@@ -56,12 +56,6 @@ final class DeepLinkAnnotatedElement {
     return annotatedElement;
   }
 
-  /** Returns the package name for the annotated TypeElement. */
-  String getPackageName() {
-    String qualifiedName = annotatedElement.getQualifiedName().toString();
-    return qualifiedName.replace("." + annotatedElement.getSimpleName(), "");
-  }
-
   String getMethod() {
     return method;
   }
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.java
@@ -23,7 +23,7 @@ import javax.lang.model.element.TypeElement;
 final class DeepLinkAnnotatedElement {
   private final String uri;
   private final DeepLinkEntry.Type annotationType;
-  private final TypeElement activityElement;
+  private final TypeElement annotatedElement;
   private final String method;
 
   DeepLinkAnnotatedElement(String annotation, Element element, DeepLinkEntry.Type type)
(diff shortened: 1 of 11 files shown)


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.java
```diff
@@ -52,8 +52,14 @@ String getUri() {
...
 
-  TypeElement getActivityElement() {
-    return activityElement;
+  TypeElement getAnnotatedElement() {
+    return annotatedElement;
+  }
+
+  /** Returns the package name for the annotated TypeElement. */
+  String getPackageName() {
+    String qualifiedName = annotatedElement.getQualifiedName().toString();
```
thenickreynolds: `javax.lang.model.util.Elements#getPackageOf` will this work instead of string parsing? I think your existing code may break with inner classes

thenickreynolds: also - you can do this in a separate class/method since this is not specific to this class

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -116,13 +113,14 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
         = !roundEnv.getElementsAnnotatedWith(DeepLinkHandler.class).isEmpty();
 
     if (!deepLinkElements.isEmpty()) {
+      String packageName = deepLinkElements.get(0).getPackageName();
```
thenickreynolds: I'm not super familiar with this code - are all of the items called in `process` from the same package? Are there possibly more than one package per `process` call?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.java
felipecsl: whoa great find! didn't know about `getPackageOf()`

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
felipecsl: yes, if you annotate classes in different packages, we'll pretty much use the package from the first one. Since I don't think order is guaranteed, we'll effectively pick a random package to use, based on all of the packages of the annotated classes 😬 
If you can think of a better solution plz let me know

seanabraham: Is the package here the package the generated code winds up winds up in?

felipecsl: yep

cdeonier:
:+1:

thenickreynolds:
Bummer about the package... Some ideas:

1. if they have a `DeepLinkHandler` defined we could use that package, more likely to be the one to put it in
2. we could try to figure out the root package based on all classes examined
3. we could also have a new annotation to define the root package (seems excessive)

This class should never be referenced by non-generated code so I guess it's acceptable.

felipecsl:
1. That's a good idea, we could do it, but only works if a `DeepLinkHandler` is defined.
1. I'm not super convinced it would work in all cases.
1. Yeah excessive.

Issue #131, opened 2016-12-13
API for checking if a URI is supported


kxfang:
I'm dealing with a legacy code base that can't be migrated over to this library all at once and would like to execute the legacy dispatch logic if we get a link that hasn't been migrated over to DeepLinkDispatch yet. Is there any interest in supporting an API that allows the client to query whether or not a URI is supported?

felipecsl:
You can use a helper method like this:

```java
public static boolean isDeepLink(Intent intent) {
  return intent.getBooleanExtra(DeepLink.IS_DEEP_LINK, false);
}
```

kxfang:
To confirm, the snippet you posted is for dispatched Activities to determine whether or not the Intent is coming from a deep link, is that correct? Our use case is a little different. What we want is something like:
```
if (DeepLinkDispatch.supportsUri(someExternalUri)) {
   UrlDispatch.dispatch(someExternalUri);
} else {
   doSomeSpecialHandling();
}
```

felipecsl:
Oh, I see what you're saying. Yeah that one is not supported yet, but should be easy to add support for. I'll try to get it in in the next few days!

kxfang:
Awesome, thank you!

felipecsl:
Closed via #133

PR #133, opened 2016-12-14
Adds support for checking if a URI is supported

MichaelEvans:
Addresses #131.

cc: @kxfang

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -313,6 +313,15 @@ public class DeepLinkProcessor extends AbstractProcessor {
         .addStatement("throw new $T($S)", AssertionError.class, "No instances.")
         .build();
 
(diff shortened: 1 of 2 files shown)


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -313,6 +313,15 @@ private void generateDeepLinkDelegate(String packageName) throws IOException {
         .addStatement("throw new $T($S)", AssertionError.class, "No instances.")
         .build();
 
+    MethodSpec supportsUri = MethodSpec.methodBuilder("supportsUri")
+        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+        .returns(TypeName.BOOLEAN)
+        .addParameter(String.class, "uriString")
```
MichaelEvans: Wasn't sure if it was better to take a `String` or a `Uri`. I don't really feel strongly either way, it'd be just as easy to call `uri.toString()` before passing it into `parseUri()`.

felipecsl:
That was fast, thanks!!

kxfang:
Awesome. Thanks all!

PR #136, opened 2016-12-16
[3.0] Add support for DeepLink modules

felipecsl:
This is a new concept on DeepLinkDispatch 3.0 and will allow for Android applications that have multiple modules to register `@DeepLink` annotations across multiple modules and have them be collected by the `@DeepLinkHandler` annotated `Activity`. 
Example:
```java
@DeepLinkHandler({ FooModule.class, BarModule.class })
public class DeepLinkActivity extends Activity {
  @Override protected void onCreate(Bundle savedInstanceState) {
    super.onCreate(savedInstanceState);
    DeepLinkDelegate deepLinkDelegate = new DeepLinkDelegate(new FooModuleLoader(), 
      new BarModuleLoader());
    deepLinkDelegate.dispatchFrom(this);
    finish();
  }
}
```

The `SampleModuleLoader.class` is generated by the Processor (formerly called `DeepLinkLoader`). Every class annotated with `@DeepLinkModule` will cause DeepLinkDispatch to generate a `*Loader` class, that can be then passed into the `DeepLinkDelegate` constructor.


MichaelEvans:
It'd be great to have a follow up with a sample of multi-module support.

MichaelEvans:
Do you need to add `@DeepLinkModule` to `getSupportedAnnotationTypes()`?

MichaelEvans:
Seems like with these changes, we've lost the ability to _not_ use `DeepLinkDelegate`? I thought it was a very compelling use case for a lot of apps to take the simple approach and just annotate activities, if they didn't need the `DeepLinkDelegate`. Do you think it's worth making this type optional again?

felipecsl:
* Added `sample-library` project
* Added `DeepLinkModule` to `getSupportedAnnotationTypes()`
* @MichaelEvans as discussed offline, we can't easily do that now because of the new `@DeepLinkModule` annotation, which is required for the loader to be generated.

jonathan-caryl:
It'd also be very helpful if there was a guide to migrating from 2.0.1 to 3.0.0, unless there is one already that I didn't see?

felipecsl:
@jonathan-caryl the README has been updated with the instructions for v3 already! Let us know if you still have questions after following that!

PR #150, opened 2017-02-03
Allow custom annotations across modules

### LINKED ISSUE #149: Custom @DeepLinkSpec annotations don't work across modules
Noticed this while integrating into our project. Doesn't block us, but someone might encounter the problem eventually.

kxfang:
Fixes #149

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -94,11 +94,7 @@ public class DeepLinkProcessor extends AbstractProcessor {
   }
 
   @Override public Set<String> getSupportedAnnotationTypes() {
-    return Sets.newHashSet(
-        DeepLink.class.getCanonicalName(),
-        DeepLinkSpec.class.getCanonicalName(),
-        DeepLinkModule.class.getCanonicalName(),
-        DeepLinkHandler.class.getCanonicalName());
+    return Sets.newHashSet("*");
   }
 
   @Override public SourceVersion getSupportedSourceVersion() {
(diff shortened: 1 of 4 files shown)


kxfang:
`roundEnv.getElementsAnnotatedWith(DEEP_LINK_SPEC_CLASS)` only picks up custom annotations from the current module, whereas we're interested in custom annotations from all modules.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -94,11 +94,7 @@
   }
 
   @Override public Set<String> getSupportedAnnotationTypes() {
-    return Sets.newHashSet(
-        DeepLink.class.getCanonicalName(),
-        DeepLinkSpec.class.getCanonicalName(),
-        DeepLinkModule.class.getCanonicalName(),
-        DeepLinkHandler.class.getCanonicalName());
+    return Sets.newHashSet("*");
```
felipecsl: What are the implications of doing this?

kxfang: I'm learning as I go here, so thanks for bearing with me –

> `public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv)`

The annotations in the parameter of the `process` method is determined by what we return from `getSupportedAnnotationTypes()`. Since there's no way we can know statically which annotations we want to look at, we end up having to examine all annotations.

Previously we were returning `true` from at the end of the `process` method, which "claims" our DeepLink annotations so other processors can't touch them. Now we must return `false`, otherwise we claim annotations owned by other processors, which would be bad. Not claiming annotations at all seems to be okay, since another well-behaved processor should just ignore annotations that it doesn't own. I looked at the Dagger processor and they also never claim any annotations.

felipecsl: Nice, TIL 😄

PR #167, opened 2017-03-16
update Annotations with a lower retention.

kaedea:
Annotation with CLASS retention will be recorded in compiled `.class` file, which is not necessary for DeepLink. 

Besides, in this condition, the Android assemble task will fail when the ProgGuard is enabled because the the annotation class may be reversed by ProGuard. The annotation info recorded in class file will not match the reserved annotation class. Check out the following message.

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:transformClassesWithMultidexlistForRelease'.
> java.lang.UnsupportedOperationException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```


felipecsl:
Seems like the unit tests failed with this change

kaedea:
Unit test has been fixed.

felipecsl:
Thanks!

PR #192, opened 2017-08-29
Ensure a sorted DeepLinkEntry registry

jingwei423:
1 add logic to ensure a stable sorting result
2 move the doc generation after the sorting, so the doc is sorted

This is important for us: with sorted outputs we could decide whether
or not to trigger different downstream processes, so we can save time
on build tasks, and commit verifications


felipecsl:
@jingwei423 can you look into the checkstyle failures? I can merge after that's green1

jingwei423:
@felipecsl thanks for the review! Can we have a new release soon? So we can update our internal tools to depend on this change.

jingwei423:
@felipecsl can we have a release for this?

felipecsl:
Sure, we'll get one out soon!

felipecsl:
v3.1.1 has been uploaded. It should be available soon

jingwei423:
Awesome, thank you!

PR #196, opened 2017-09-06
Version bump for v3.1.1

felipecsl:


felipecsl:
Actually we can keep 3.2.0 for the snapshot... Closing

PR #221, opened 2018-06-19
Create BaseDeeplinkDelegate

BenSchwab:
## What
Currently, in multi-module projects DLD generated a `DeepLinkDelegate` class which knows about all the DeepLink modules.

This means that the DeepLinkDelegate is only accessible from the application module, and is hard to use in feature modules without managing some interface based binding.

This PR opens up access to a `BaseDeepLinkDelegate` which `DeepLinkDelegate` extends from. The BaseDeepLinkDelegate can be initialized from a list of `Parsers`. This is useful, because you can use [multibinding](https://google.github.io/dagger/multibindings.html) to bind the DeepLink loaders in various feature modules, but keep the definition of the `DeepLinkDelegate` in a base/core module.

## How
Currently, DLD generates almost all it’s code via an annotation processor. However, almost all of this code (besides creating the constructor which knows about all the DeepLinkDelegates) is static.
This PR moves the static code the the aforementioned `BaseDeepLinkDelegate` which `DeepLinkDelegate` extends from. Thus *this PR is backwards compatible with previous versions of DLD*. 

I think moving code to java source vs annotation processor nice because future DLD changes won't have to edit the annotation processor for static code changes.

However, the DLD runtime did not have to depend on Android before, which now it does as the non-generated code references Android code. 

So this PR makes `deeplink-dispatch` depend on android, and moves the non-android dependent portion to `deeplink-dispatch-base`.

Lastly, this PR DRYs  Gradle dependency management as I had to create a new module and did not want to repeat all the dependencies.

@felipecsl @gpeal @rossbacher


deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -0,0 +1,164 @@
...
+import android.support.v4.content.LocalBroadcastManager;
+import android.util.Log;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings({"WeakerAccess", "unused"})
+public class BaseDeepLinkDelegate {
```
BenSchwab: All of this was copied from the generated code.

build.gradle
```diff
@@ -1,23 +1,39 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+  apply from: rootProject.file("dependencies.gradle")
+
   repositories {
+    google()
     jcenter()
     google()
```
rossbacher: Remove this? You already have it as first.

build.gradle
```diff
@@ -1,23 +1,39 @@
...
   }
 }
 
 allprojects {
+  apply from: rootProject.file("dependencies.gradle")
+
   repositories {
+    google()
     jcenter()
     google()
```
rossbacher: Remove this? You already have it as first.

deeplinkdispatch-processor/build.gradle
```diff
@@ -5,20 +5,29 @@ sourceCompatibility = 1.7
...
+}
+
 dependencies {
-  compile project(':deeplinkdispatch')
-  compile 'com.google.auto.service:auto-service:1.0-rc2'
-  compile 'com.squareup:javapoet:1.8.0'
-  compile 'com.google.auto:auto-common:0.6'
+  implementation project(':deeplinkdispatchbase')
+  implementation 'com.google.auto.service:auto-service:1.0-rc3'
+  implementation deps.jsr305
```
rossbacher: I don't see any direct dependency on jsr305.

rossbacher: I thought we did not have it before but we did, the dependency was just moved.

deeplinkdispatch-processor/build.gradle
```diff
@@ -5,20 +5,29 @@ sourceCompatibility = 1.7
 apply plugin: 'com.bmuschko.nexus'
 apply plugin: 'checkstyle'
 
+sourceSets {
+  test {
+    java {
+      srcDir '../deeplinkdispatch/src/main/java' // In tests we need access to the runtime files from the dld library
```
felipecsl: why not add a regular `testImplementation` dependency on that?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  public DeepLinkResult dispatchFrom(Activity activity) {
+    if (activity == null) {
+      throw new NullPointerException("activity == null");
+    }
+    return dispatchFrom(activity, activity.getIntent());
+  }
+
+  public DeepLinkResult dispatchFrom(Activity activity, Intent sourceIntent) {
```
felipecsl: this is a nice opportunity for us to clean up/break down this method. Doesnt have to happen on this PR

dependencies.gradle
```diff
@@ -0,0 +1,44 @@
+def versions = [
+    kotlinVersion                : '1.2.21',
```
felipecsl: might want to bump the kotlin version?

dependencies.gradle
```diff
@@ -0,0 +1,44 @@
...
+    supportLibVersion            : '27.1.1',
+    mockito                      : '1.9.5',
+    testRunner                   : '0.5',
+    espressoIdlingResourceVersion: '2.2.2',
+    roboelectricVersion          : '3.5',
+    spekVersion                  : '1.1.5',
```
felipecsl: this doesnt seem to be used anywhere

settings.gradle
```diff
@@ -1 +1 @@
-include ':sample', ':deeplinkdispatch', ':deeplinkdispatch-processor', ':sample-library'
+include ':sample', ':deeplinkdispatch',':deeplinkdispatchbase', ':deeplinkdispatch-processor', ':sample-library'
```
felipecsl: nit: i'd rename to `deeplinkdispatch-base` just for readability/consistency

deeplinkdispatch-processor/build.gradle
BenSchwab: The processor is java module, and the `deeplinkdispatch` is now an Android module. Gradle doesn't seem to allow a java module to depend on an Android module. I don't love copying the class, but it got the job done. If the runtime component gets any more complicated, we should think of a different solution.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
BenSchwab: Agreed 👍 I have some requests for some new features that will be easier to work on now that I can just edit a java file.

deeplinkdispatch-processor/build.gradle
felipecsl: ahh good point, thats a limitation.
you can do something like this instead then:
```
testImplementation project(':deeplinkdispatch').sourceSets.main.output
```
it should be a bit less fragile

dependencies.gradle
BenSchwab: I'll leave this to a follow-up, partially because buck currently has some issues with `1.2.25`

deeplinkdispatch-processor/build.gradle
BenSchwab: I tried your snippet and  `testImplementation project(':deeplinkdispatch').android.sourceSets.main.output`

But seems like android projects don't package output the same way. If you have any ideas I can follow up.

PR #231, opened 2018-09-26
Fix Travis YML file

rogerhu:
Add default license to avoid needing to use SDK Manager to accept

Taken from: https://chromium.googlesource.com/android_tools/+/refs/heads/master/sdk/licenses/android-sdk-license


rogerhu:
anyone home? :)

PR #240, opened 2019-01-29
Update maven release pipeline

rossbacher:
- Updated the maven release pipeline to work with new `.aar` and `.jar` mixed release
- Removed Nexus and Android Maven plugin in favor of https://github.com/airbnb/epoxy/blob/master/gradle/gradle-maven-push.gradle
- Moved POM generation into `gradle-maven-push.gradle`.
- Moved all common POM values to root `gradle.properties` and all non common ones to projects `gradle.properties` files.


gradle-maven-push.gradle
```diff
@@ -0,0 +1,220 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'maven'
+apply plugin: 'signing'
+
+version = VERSION_NAME
```
mmdango: @rossbacher do you mean `PROJECT_VERSION`? Also you set `version` and `group`, but don't seem to use them

rossbacher: @mmdango Yes, I created a merge conflict here. I just pushed a fix.

rossbacher: @mmdango Does not look like it, I'm removing them.

PR #260, opened 2019-10-03
Expose Parsers and their Registries of Deep Links

adellhk:
To be able to access all deeplinks captured by DLD,
without needing reflection, add getRegistry() to Parser and
change Parser to an Abstract Class.

Update tests to match.

To be able to access a destination, add ComponentName to
DeepLinkResult.

Potentially, storing more information, like Extras could
allow for unit tests to assert that URI parameters have
been parsed as extras. To enable passing Android-framework
things in DeepLinkResult, @BenSchwab pointed out that it
may be possible to move DeepLinkResult to an Android
module (out of deeplinkdispatch-base).


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
```diff
@@ -7,11 +7,14 @@
   private final boolean successful;
   private final String error;
   @Nullable private final String uriString;
+  private final String componentName;
 
-  public DeepLinkResult(boolean successful, @Nullable String uriString, String error) {
+  public DeepLinkResult(boolean successful, @Nullable String uriString, String error,
```
elihart: would be nice to have a comment explaining what the format of componentName is

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
```diff
@@ -29,6 +32,11 @@ public String error() {
     return error;
   }
 
+  /** @return the name of the component which is the DeepLink's destination. */
```
elihart: `name of the component` isn't very helpful. Is this the fully qualified class name? can you give an example?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
```diff
@@ -1,8 +1,35 @@
 package com.airbnb.deeplinkdispatch;
 
+import java.util.List;
+
 /**
  * Parser classes are used for keeping a registry of deep links and dispatching them from Intents.
  */
-public interface Parser {
-  DeepLinkEntry parseUri(String uri);
+public abstract class Parser {
```
elihart: this could be a good time to introduce kotlin to this project and make this a kotlin class. that would also help make clear that this is a read only list

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
```diff
@@ -1,8 +1,35 @@
...
+  }
+
+  /**
+   * The List of DeepLinkEntries collected from a Module. Annotation Processing-generated Children
+   * of this class will pass the collected {@code List<DeepLinkEntry>} in the constructor's
+   * {@code super()}.
+   */
+  private final List<DeepLinkEntry> registry;
+
+  public List<DeepLinkEntry> getRegistry() {
```
elihart: should explain that this list should not be modified via javadoc?

elihart: registry is a bit of an unusual name for a list. it isn't clear based on this class why that name is used

adellhk: Are you thinking of something more explicit than the comment at the top of this file? https://github.com/airbnb/DeepLinkDispatch/pull/260/files/256268ae06907ec3dcb220ace44280a99cc7e624#diff-54c0eadfe4fe7515a50210ca4ed3ceecR6

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
adellhk: That's true, but I don't want to do that right now.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
adellhk: What do you mean by format? Do you mean something beyond https://github.com/airbnb/DeepLinkDispatch/pull/260/files/256268ae06907ec3dcb220ace44280a99cc7e624#diff-0616d20e785ed562a4ad36be4f682e4dR35

elihart: yes, "name" is vague. There is simple name, qualified name, canonical name, etc. It's pretty safe to assume that FQN name is what is meant here, but it should be explicit

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
elihart: A clearer name would be something like `getRegisteredDeeplinks`. you just generally don't see a generic collection referred to as a specific name like this

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
adellhk: Okay. Do you think it makes sense to duplicate the comment on [L36](https://github.com/airbnb/DeepLinkDispatch/pull/260/files#diff-0616d20e785ed562a4ad36be4f682e4dR36) here or to delete that one and move it here?

adellhk: I decided to duplicate it across both the property and the getter^

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
```diff
@@ -1,8 +1,40 @@
...
+
+  /**
+   * Notice that {@link com.airbnb.deeplinkdispatch.Parser#registry} is immutable.
+   * @return {@code Collections.unmodifiableList<DeepLinkEntry>} {@link Parser#registry}
+   */
+  public List<DeepLinkEntry> getRegisteredDeepLinks() {
+    return registry;
+  }
+
+  public DeepLinkEntry parseUri(String uri) {
```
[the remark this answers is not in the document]
adellhk: Some thoughts:

- This seems like something we should address when [generating the list](com.airbnb.deeplinkdispatch.DeepLinkProcessor#generateDeepLinkLoader), not when retrieving elements from it. 
  - If we wanted to prevent **intra**-module duplicates we could fail at compile-time by adding some checks, using an immutable Set (instead of list, `Collections.unmodifiableSet`), or both.
  - I'm not sure if app-wide duplicates are a possibility, but we could also check for this.
- This doesn't change the prior behavior, which would have generated code identical code like in the [tests](https://github.com/airbnb/DeepLinkDispatch/pull/260/files/7085625781a2b64e8b1c429e96d40d7aa265bf5f#diff-49396e9f26b9dc5c42b6266107b3832aL75).
- I don't want to tackle additional scope in this PR.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
```diff
@@ -1,8 +1,40 @@
...
+    this.registry = registry;
+  }
+
+  /**
+   * The List of DeepLinkEntries collected from a Module. Annotation Processing-generated Children
+   * of this class will pass the collected {@code List<DeepLinkEntry>} in the constructor's
+   * {@code super()}. The generated implementation will be a {@link
+   * Collections#unmodifiableList(java.util.List)}.
+   */
+  private final List<DeepLinkEntry> registry;
```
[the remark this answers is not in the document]
adellhk: We lint on Circle. I will change this.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
```diff
@@ -7,11 +7,19 @@
   private final boolean successful;
   private final String error;
   @Nullable private final String uriString;
+  /**
+   * The fully-qualified name of the {@link android.content.ComponentName} component
+   * which is the DeepLink's destination. eg. {@code
+   * com.airbnb.deeplinkdispatch.sample.DeepLinkActivity}
+   */
+  private final String componentName;
```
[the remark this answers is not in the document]
adellhk: I didn't think there was any additional useful information in `Class` beyond the fully-qualified name.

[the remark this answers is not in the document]
rossbacher: Well if it has a type it would be clear what it is without having to read the documentation. Is there a reason it is not a `Class`?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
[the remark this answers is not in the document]
rossbacher: Maybe we should create a ticket . to track this?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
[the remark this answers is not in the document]
adellhk: Just to be smaller/store less information.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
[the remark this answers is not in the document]
adellhk: https://github.com/airbnb/DeepLinkDispatch/issues/261

gradle-maven-push.gradle
```diff
@@ -213,7 +213,8 @@ afterEvaluate { project ->
         if (project.getPlugins().hasPlugin('com.android.application') ||
                 project.getPlugins().hasPlugin('com.android.library')) {
             archives androidSourcesJar
-            archives androidJavadocsJar
+            // todo AdellHk - javadocs classpath is broken eg. BaseDeepLinkDelegate.java:8: error: package androidx.core.app does not exist
```
elihart: I would not merge with this change - Javadoc is important and we should figure out why it isn't working.

A good fix may be to migrate away from the antiquated gradle script, and use a plugin like https://github.com/vanniktech/gradle-maven-publish-plugin

rossbacher: I also think you might not be able to publish to Sonatyope with missing Javadoc.

I think the problem here is not the publishing, but the actual class path for the JavaDoc.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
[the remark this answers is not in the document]
rossbacher: @BenSchwab do you have an opinion here, I think this should be a `Class` type if it can be.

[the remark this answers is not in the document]
BenSchwab: My questions was if this was always an activity? I believe DLD will call startActivity, so other components wouldn't work. If so I would name this `actiivtyComponentName`

From an API perspective, I think the choice is between `ComponentName` and `String`. I would vote for `ComponentName` as that is the official class which represents the piece of information that we are modeling. The choice to use FQN is one that was made for the use-case we had in mind, but at this infrastructure level we shouldn't design for that, imo.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/Parser.java
```diff
@@ -1,8 +1,41 @@
...
+  /**
+   * Notice that {@link com.airbnb.deeplinkdispatch.Parser#registry} is immutable.
+   *
+   * @return {@code Collections.unmodifiableList<DeepLinkEntry>} {@link Parser#registry}
+   */
+  public List<DeepLinkEntry> getRegisteredDeepLinks() {
+    return registry;
+  }
+
+  public DeepLinkEntry parseUri(String uri) {
```
BenSchwab: Thanks for the other javadocs in this PR! A javadoc would be nice here too.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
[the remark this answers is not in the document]
BenSchwab: I wonder reading this again if it should actually be an intent.. That really is what the information of result is, and might be a nice step in decoupling matching from launching

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -130,28 +134,35 @@ public DeepLinkResult dispatchFrom(Activity activity, Intent sourceIntent) {
...
+                newIntent.getComponent().toString());
       } catch (NoSuchMethodException exception) {
         return createResultAndNotify(activity, false, uri,
-                entry.getUriTemplate(), "Deep link to non-existent method: " + entry.getMethod());
+                entry.getUriTemplate(), "Deep link to non-existent method: " + entry.getMethod(),
+               "");
       } catch (IllegalAccessException exception) {
         return createResultAndNotify(activity, false, uri,
-                entry.getUriTemplate(), "Could not deep link to method: " + entry.getMethod());
+                entry.getUriTemplate(), "Could not deep link to method: " + entry.getMethod(),
```
BenSchwab: Instead of the string for the error, we could pass the exception. (Again, that is the most accurate representation of the result.)

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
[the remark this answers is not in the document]
adellhk: This is resolved with ben's help today :)

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkResult.java
```diff
@@ -0,0 +1,112 @@
+package com.airbnb.deeplinkdispatch;
+
+import android.content.Intent;
+import androidx.core.app.TaskStackBuilder;
+
+import javax.annotation.Nullable;
+
+@SuppressWarnings({"unused", "WeakerAccess"})
+public final class DeepLinkResult {
```
elihart: imo we should really add kotlin to this project and and use kotlin for all new code - continuing with Java is tech debt.

adellhk: @elihart 

- This file is just green because I moved `DeepLinkResult` from module `deeplinkdispatch-base` to module `deeplinkdispatch`.
- Do you think this [principle from google](https://google.github.io/eng-practices/review/reviewer/standard.html) applies?

> In general, reviewers should favor approving a CL once it is in a state where it definitely improves the overall code health of the system being worked on, even if the CL isn’t perfect.

adellhk:
Currently blocked from merging this based on androidJavadocs task failing.

elihart:
@adellhk I would recommend deleting the `gradle-maven-push.gradle` approach and switching to an approach like https://github.com/airbnb/MvRx/blob/master/build.gradle#L33

This is what we've moved to for our other open source projects - it is simpler and works better

adellhk:
@elihart oh yes, thanks! I forgot about your recommendation. I'll investigate this tomorrow.

adellhk:
Now with the vanniktech's Gradle-maven-publish-plugin and javadocs!
Screen Shot 2019-11-15 at 3 52 53 PM

dependencies.gradle
```diff
@@ -23,6 +23,7 @@ ext.deps = [
     javaPoet                 : 'com.squareup:javapoet:1.9.0',
     jsr305                   : 'com.google.code.findbugs:jsr305:3.0.2',
     okio                     : 'com.squareup.okio:okio:1.13.0',
+    gradleMavenPublishPlugin : 'com.vanniktech:gradle-maven-publish-plugin:0.8.0',
```
elihart: @adellhk unless you use the 0.9.0 snapshot it won't properly include kotlin sources. This is documented in the mvrx setup that I linked you too.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Oct 23 13:41:07 PDT 2019
+#Mon Nov 18 16:19:47 PST 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6-all.zip
```
elihart: might as well use 5.6.4?

dependencies.gradle
```diff
@@ -6,9 +6,9 @@ def versions = [
 
 ext.versions = versions
 ext.androidConfig = [
-    agpVersion       : '3.2.0',
+    agpVersion       : '3.3.0',
```
elihart: we should get this updated to 3.5.0 in a follow up PR

PR #262, opened 2019-10-23
Use Kotlin Plugins for jdk 1.7

adellhk:
@elihart 

Also updated BuildToolsVersion because of this message Screen Shot 2019-10-23 at 11 28 37 AM


dependencies.gradle
```diff
@@ -8,7 +8,7 @@ ext.versions = versions
 ext.androidConfig = [
     agpVersion       : '3.2.0',
```
elihart: should update agp while you're at it

PR #265, opened 2019-11-14
improve performance of finding entry

bobby1i:
If we have lots of entries in a parser, it would take long time to init and parse.

1. init regex when really using DeepLinkEntry
2. match prefix first in order to avoid matching regex for every entry

I write a test case which contains 10,000 entries, and it reduce process time from 1s to ~80ms on my macbook pro with i9 core. And I think it would improve more on phone.

@adellhk @BenSchwab @elihart @rossbacher


adellhk:
Hey @bobby1i , this looks great! I will wait for others to chime in. It seems like your improvements grant ~1 order of magnitude improvement! With changes to `DeepLinkEntry`, Rossi and I can maybe get another order of magnitude, but it may also be unnecessary.

Some additional context: @rossbacher and I have some time scoped for this quarter to address deep link runtime performance and some tasks related to multi-app developer experience.

bobby1i:
Agree with @elihart , let me migrate DeepLinkEntry to kotlin

bobby1i:
I've migrated DeepLinkEntry to kotlin, and it gets slower. 10,000 entries parser init & finding time increases from ~80ms to ~200ms.

I think it's due to performance of some kotlin method. Do you guys think is it worth that we keep it in JAVA.

adellhk:
@bobby1i  oh weird 🤔. Would you push the slower kotlin code you experienced to this branch in a commit? If you haven't thought of it already (this is what I'll do), you can decompile it to Java to see why it's slower.

elihart:
I would guess that `lazy` slows it down. This is by default synchronized. You could use the unsynchronized version, which is less safe but equivalent to the java version.

happy to take a look if you push the kotlin code too. I think a slightly slower time than Java is worth the safety and maintainability (still a lot better than 1 second)

rossbacher:
Actually we need/want to rewrite this completely anyway. I'd be ok to not convert this to Kotlin now and just do this enhancement.

bobby1i:
I've pushed the converted DeepLinkEntry. I'm sure that the little performance loss is due to lazy pattern and kotlin non-null checking, and in my second test the parser init & finding time is from ~80ms to ~120ms. Since it's still fast enough, I'd like to use the kotlin one.
@adellhk @elihart @rossbacher PTAL, especially the new DeepLinkEntry class

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -0,0 +1,106 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.airbnb.deeplinkdispatch
+
+import java.util.regex.Pattern
+
+class DeepLinkEntry(
```
elihart: this should maybe be a data class - having proper `toString` for debugging and printing would be nice

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -0,0 +1,106 @@
...
+    val uriTemplate: String,
+    val type: Type,
+    /**
+     * The class where the annotation corresponding to where an instance of DeepLinkEntry is declared.
+     */
+    val activityClass: Class<*>,
+    val method: String? = null
+) {
+  private val prefix: String by lazy(mode = LazyThreadSafetyMode.NONE) {
```
elihart: it would be worth documenting here why we don't use thread safe (ie for performance reasons and because we don't expect parallel usage)

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -0,0 +1,106 @@
...
+   * @return the map of parameter values, where all values will be strings.
+   */
+  fun getParameters(inputUri: String): MutableMap<String, String> {
+    val paramsMap = mutableMapOf<String, String>()
+    val deepLinkUri = DeepLinkUri.parse(inputUri)
+    regex.matcher(deepLinkUri.schemeHostAndPath()).takeIf {
+      it.matches()
+    }?.let { matcher ->
+      var i = 1
+      parameters.forEach { key ->
```
elihart: can you use `forEachIndexed { index, key ->` and get rid of `i`?

bobby1i:
comment addressed. @adellhk could you help to release this? I have no access.

rossbacher:
You might get different results are the search result is in a different part of the ModuleLoader.
With this setup the creation of the patterns is lazy and the fewer you have to do the faster it is.
I'm rewriting this fully and I think we can get bigger gains. I will include the prefix idea into my change too, for now I would say not not merge it.

Issue #247, opened 2019-04-02
Incremental annotation processing


markchristopherng:
Hi,

Does this project support incremental annotation processing? If not are there any plans to support it to help improve build speed.

markchristopherng:
@elihart I am just wondering if you will have time to look into this, since you mentioned that it might not require a code change?

kraghu:
Has any one have a solution to this ? We are unable to update our libraries because of this

marcosalis:
Anything new on the above? We're unfortunately blocked on using incremental annotation processing with `kapt` because of the library.
Thank you! 🙏

radityagumay:
Hi any updates on this?

elihart:
When there is an update it will be shared.

In the meantime, feel free to contribute a PR to add incremental support if you want it. It should be fairly straightforward - I believe this would simply be an aggregating processor

benjamin-bader:
I just took a stab at this, and I don't believe it will be simple.  The issue is that aggregating processors may only support annotations with `CLASS` or `RUNTIME` retention.  Any processors declaring themselves to be aggregating, and which declare support for `SOURCE` annotations, will cause a build failure.

That's the case with the `DeepLinkProcessor`.  It declares support for `"*"`, which means every annotation on the classpath.  I assume this is to support `@DeepLinkSpec`.  I'm not sure how it will be possible to support arbitrary "qualified" spec annotations and still be an aggregating processor.

It seems like it _should_ be possible; I think dagger-compiler is an incremental processor, but I've no idea how they manage it.  Maybe their processors just don't "consume" `@Qualifier`-annotated things - if so, they're of no help here.

LachlanMcKee:
I believe I have a solution for making the processor incremental (aggregating): https://github.com/airbnb/DeepLinkDispatch/pull/266

This would require users to specify their custom annotations within the build.gradle. Currently I have made it mandatory in the PR, but it could be an optional feature that allowed developers to opt-in to incremental processing.

LachlanMcKee:
In case anyone is interested, I have compiled the binaries on Jitpack (hopefully that's okay). I've tested it out in my current project, and it works a treat :)

```gradle
com.github.LachlanMcKee.DeepLinkDispatch:deeplinkdispatch:4.3.0-INCREMENTAL
com.github.LachlanMcKee.DeepLinkDispatch:deeplinkdispatch-processor:4.3.0-INCREMENTAL
```

[...]

PR #266, opened 2019-11-18
Made processor 'aggregating' incremental

LachlanMcKee:
Hi!

I have managed to make the annotation processor incremental (as suggested via #247)

The main issue that was encountered by @benjamin-bader was that the custom annotations would not work, as you need to be explicit about each annotation that the processor can handle. I believe Qualifiers are different in dagger, as they are used on methods, not classes.

The trick I have used is to use an annotation processor option to allow the user to specify each custom annotation, and then the processor can work as expected.

The obvious downside to this is that users have to remember to do this. I have added some error handling that should remind them.

Thoughts? I don't know if there is a more elegant way to do this.

@adellhk @BenSchwab @elihart @rossbacher


LachlanMcKee:
This could potentially be an opt-in feature, as you could mark the processor as dynamic, and only make it aggregating if the user decides to specify the annotations in the build.gradle?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -104,6 +114,20 @@
 
   @Override
   public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+    try {
+      processInternal(annotations, roundEnv);
+    } catch (DeepLinkProcessorException e) {
+      error(e.getElement(), e.getMessage());
+    }
+    return true;
```
elihart: why did you change this to return true? I think in general it is best to return false so other processors can use the annotations too

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkHandler.java
```diff
@@ -26,7 +26,7 @@
  * incoming {@code Intent} to the correct Activities annotated with {@link DeepLink}.
  */
 @Target({ ElementType.TYPE })
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.CLASS)
```
elihart: seems like we should add a check somewhere that all annotations for `DeepLinkSpec` should also have non source retention

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -104,6 +114,20 @@
...
+  }
+
+  private void processInternal(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+    if (getCustomAnnotations().length == 0) {
```
elihart: it should be fine to have this be empty if no custom annotations have been defined. 

Can we move this check to line 134, and when adding a custom annotation element make sure the type is already defined in `getCustomAnnotations()`?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -90,8 +87,21 @@
     documentor = new Documentor(processingEnv);
   }
 
+  private String[] getCustomAnnotations() {
```
elihart: ideally this would be something like a lazy val in kotlin so we don't have to recompute it on each access

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -229,7 +251,19 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
     for (AnnotationMirror customAnnotation : annotationMirrors) {
       List<? extends AnnotationValue> suffixes =
           asAnnotationValues(AnnotationMirrors.getAnnotationValue(customAnnotation, "value"));
-      String[] prefixes = prefixesMap.get(customAnnotation.getAnnotationType().asElement());
+
+      Element customElement = customAnnotation.getAnnotationType().asElement();
```
elihart: it's not clear to me what this has to do with the annotation processing change

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessorException.java
```diff
@@ -0,0 +1,20 @@
+package com.airbnb.deeplinkdispatch;
+
+import javax.lang.model.element.Element;
+
+class DeepLinkProcessorException extends RuntimeException {
```
elihart: this should be in kotlin. https://github.com/airbnb/DeepLinkDispatch/pull/260 is adding kotlin dependency

deeplinkdispatch-processor/build.gradle
```diff
@@ -17,7 +17,6 @@ sourceSets {
 
 dependencies {
   implementation project(':deeplinkdispatch-base')
-  implementation 'com.google.auto.service:auto-service:1.0-rc3'
```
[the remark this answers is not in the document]
benjamin-bader: Given that the right META-INF files are already added here, there's not much reason to keep the dependency.  In my experience (author of https://github.com/keepsafe/dexcount-gradle-plugin) these files are among the most stable part of the codebase - I don't think I've ever needed to change them.  IMO, AutoService doesn't really pay for itself unless you're also making use of `BasicAnnotationProcessor`, etc.

If you decide you _do_ want to keep this, I'd suggest switching this dependency to `implementation 'com.google.auto.service:auto-service-annotations:1.0-rc6'`, and adding another on `annotationProcessor 'com.google.auto.service:auto-service:1.0-rc6'`, to keep the runtime classpath cleaner.  (I've noticed that AutoService seems to spill over from this project into ours because of the current configuration.  I'm confused by this, but when I modified the deeplink processor locally this way, AutoService was no longer being run in our project.)

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
LachlanMcKee: A fair point. I'll change it back.

My original reasoning to set this to false is that if you make your processor incremental (isolating), and another processor relies on your annotation as well, it could potentially rerun your processor as well.

But considering this is an aggregating processor, it doesn't really matter.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkHandler.java
LachlanMcKee: Gradle actually does this for you. If you reference any classes that use `SOURCE` retention in `getSupportedAnnotationTypes()`, then Gradle will fail when attempting to run the processor:

```
> Task :sample-library:compileDebugJavaWithJavac FAILED
error: '@DeepLinkModule' has source retention. Aggregating annotation processors require class or runtime retention.
1 error

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sample-library:compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.
```

It might not be worth adding an additional check?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessorException.java
LachlanMcKee: done

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
LachlanMcKee: If someone does not specify the annotation in the build gradle, the annotation will not exist with the map. Without this code change, the processor throws an extremely unhelpful NullPointerException.

This is the point at which the processor is expecting to find the annotation, so it made sense to put the check here to give developers a helpful error telling them they need to add their annotation to the build.gradle

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
LachlanMcKee: By the way, just to clarify with how the incremental processor works (took me ages to get my head around it). You need to be specific about what annotations your processor handles. if you do not specify an annotation, the processor will never be told about it. 

Therefore the `annotation.getAnnotation(DEEP_LINK_SPEC_CLASS)` code would always return a null value.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
LachlanMcKee: I've created a variable for it and assigned it within `init`. I think it would be better for the whole class to be in Kotlin before attempting to make it a lazy val. Unless you have a more elegant suggestion?

LachlanMcKee:
@elihart I believe I have made the changes that you requested. Please let me know if I missed anything!

@rossbacher @adellhk happy to work with you both to get this across the line :)

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
LachlanMcKee: Ahhh. I added a test for this scenario now and have found what causes the issue...

Within the code block  `enumerateCustomDeepLinks` it looks for all annotations that meet the criteria (this is able to circumvent the incremental processor only passing back valid annotations)

This only happens when a class is annotated by something else that the annotation processor picks up. i.e.

```
@DeepLink({ "/view_users" })
@WebDeepLink({ "/users", "/user/{id}" })
public class CustomPrefixesActivity {}
```

Even though the WebDeepLink is not registered, the @DeepLink type is. Therefore the class is picked up by the process, it attempts to then find WebDeepLink in the accepted list of annotations, then crashes because it's not there.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -163,9 +190,7 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
       if (deepLinkAnnotation != null) {
         deepLinks.addAll(Arrays.asList(deepLinkAnnotation.value()));
       }
-      if (customAnnotatedElements.contains(element)) {
-        deepLinks.addAll(enumerateCustomDeepLinks(element, prefixes));
-      }
+      deepLinks.addAll(enumerateCustomDeepLinks(element, prefixes));
```
LachlanMcKee: By removing this check, it's more likely that the user will receive the message stating that their annotation has not been registered.

deeplinkdispatch-processor/build.gradle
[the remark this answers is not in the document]
elihart: Those are good points - I would also be fine with using explicit `META-INF` files like you had. I don't feel too strongly.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
elihart: this seems good, thanks!

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkHandler.java
elihart: sure, sounds like relying on the gradle error would be simpler, thanks for the details!

deeplinkdispatch-processor/build.gradle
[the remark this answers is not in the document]
LachlanMcKee: I have changed it back to using META-INF when I rebased with master. Seems a lot less likely to break going forward, and is easy to manage in a project with a single annotation processor.

IntelliJ also is quite smart about formatting the file, and if the file is wrong, the compiler unit tests will fail

deeplinkdispatch-processor/src/test/java/com/airbnb/deeplinkdispatch/DeepLinkProcessorTest.java
```diff
@@ -334,10 +343,36 @@
...
         .that(sampleActivity)
+        .withCompilerOptions("-AdeepLink.customAnnotations=com.example.MyDeepLink")
         .processedWith(new DeepLinkProcessor())
         .failsToCompile()
         .withErrorContaining(
             "Only `Intent` or `androidx.core.app.TaskStackBuilder` are supported."
                 + " Please double check your imports and try again.");
   }
+
+  @Test public void testCustomAnnotationMissingFromCompilerOptionsErrorMessage() {
```
elihart: thanks for the test

LachlanMcKee:
Hi @elihart, just a heads up - I have made a change to the PR. https://github.com/airbnb/DeepLinkDispatch/pull/266/commits/1a0ffe420f394d85de40aadc7941c9e2f95a4420

I believe it is a very good idea to make the incremental processor opt-in for now. If users upgrade and forget to set the `deepLink.customAnnotations`, it will fail silently in some circumstances. By making the incremental flag opt in, it would at least encourage them to investigate whether they need to make any more changes.

Thoughts?

elihart:
Great idea about the opt in, thanks for that change.

LachlanMcKee:
@elihart what is the process around getting this merged? I think this would be worth a release in itself!

elihart:
@adellhk @rossbacher Can you approve this and figure out a release plan?

LachlanMcKee:
In case anyone is interested, I have compiled the binaries on Jitpack (hopefully that's okay). I've tested it out in my current project, and it works a treat :)

com.github.LachlanMcKee.DeepLinkDispatch:deeplinkdispatch:4.3.0-INCREMENTAL
com.github.LachlanMcKee.DeepLinkDispatch:deeplinkdispatch-processor:4.3.0-INCREMENTAL

sample-library/build.gradle
```diff
@@ -16,3 +16,9 @@ dependencies {
   implementation deps.appCompat
   testImplementation deps.junit
 }
+
+tasks.withType(JavaCompile) {
+  options.compilerArgs << "-AdeepLink.incremental=true"
```
rossbacher: This is not in line with the documentation added in the `README.md` any reason?

LachlanMcKee: I have changed it to the same format as the `READ.md`. Turns out gradle has many different ways to do this :)

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -82,28 +81,64 @@
...
   @Override public synchronized void init(ProcessingEnvironment processingEnv) {
     super.init(processingEnv);
     filer = processingEnv.getFiler();
     messager = processingEnv.getMessager();
     documentor = new Documentor(processingEnv);
+    incrementalMetadata = getIncrementalMetadata();
+  }
+
+  private IncrementalMetadata getIncrementalMetadata() {
+    if (!"true".equals(processingEnv.getOptions().get(OPTION_INCREMENTAL))) {
```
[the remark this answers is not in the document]
LachlanMcKee: I have changed this to `equalsIgnoreCase`

LachlanMcKee:
@rossbacher I have addressed your comments

sample-library/build.gradle
rossbacher: Yeo, should just be consistent. :-)

rossbacher:
Thank you so much for doing this!

condesales:
any ideas when a new release will be created containing this?

rossbacher:
@condesales I just released `5.0.0-beta01` please give it a try. We decided to go beta with this as we totally rewrote the matcher and as it is working totally different we don't know if we covered all existing use cases. It is passing all existing tests, but be are trying to be safe with this.

LachlanMcKee:
@rossbacher I'm testing 5.1.0 in my companies app, and it appears to be working well! Thanks for getting the release sorted :)

PR #268, opened 2019-12-12
New Indexed search for Deeplink Match

rossbacher:
This update the processor to get rid of Regex and instead generate a byte array from the the list of to be handled deeplinks at compile time and use that byte array during runtime to to the lookup.

Internal tests have shown that this is about 2 orders of magnitudes faster than the current approach and also requires a fraction of the code size currently required (in our test case the amount of code executed (via trace file) went from ~70MB to under 700kb.

To achieve this we encode all supported URLs for a given module into a tree data structure but use url segments instead of characters. 
This in turn is then encoded into a byte array in the following format:
```
 1 byte  2 bytes       4 bytes                     2 bytes       n bytes       n bytes
+------+-------------+---------------------------+-------------+-------------+--------------------------------------------------------------------------------------------------------+
|      |             |                           |             |             |                                                                                                        |
| Type |Value length | Children length           | Match idx   | Value       | Children                                                                                               |
|      |             |                           |             |             |                                                                                                        |
+------+-------------+---------------------------+-------------+-------------+--------------------------------------------------------------------------------------------------------+

<----------------------+  9 byte header  +--------------------->             +------+-------------+---------------------------+-------------+-------------+---------------------------+
                                                                             |      |             |                           |             |             |                           |
                                                                             | Type |Value length | Children length           | Match idx   | Value       | Children                  |
                                                                             |      |             |                           |             |             |                           |
                                                                             +------+-------------+---------------------------+-------------+-------------+---------------------------+

```

The benefit is that we can with very cheap compare operations walk this byte array and do a lookup without having to generate an actual data structure, which would slow down init time.
As this usually runs at app start (app started via deep link) this would slow down overall execution time.


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/BaseLoader.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Use a binary match index to match the given URL. Detauls can be find in {@link MatchIndex}.
+   *
+   * @param deepLinkUri The input {@link DeepLinkUri} to match
+   * @return A {@link DeepLinkEntry} if a match was found or null if not.
+   */
+  public @Nullable
+  DeepLinkEntry idxMatch(@NonNull DeepLinkUri deepLinkUri) {
+    if (deepLinkUri == null) {
```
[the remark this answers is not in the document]
rossbacher: This is for legacy reasons, as this is no runtime guarantee as it Kotlin.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/BaseLoader.java
```diff
@@ -0,0 +1,67 @@
...
+import com.airbnb.deeplinkdispatch.base.MatchIndex;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * BaseLoader classes are used for keeping a registry of deep links and dispatching them
+ * from Intents.
+ */
+public abstract class BaseLoader {
```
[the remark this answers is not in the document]
rossbacher: I think we can do that, there is no reason for this IMHO.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.airbnb.deeplinkdispatch
+
+class DeepLinkEntry(val uriTemplate: String, val type: Type,
```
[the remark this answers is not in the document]
rossbacher: I think i did format this but let me look into this.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -0,0 +1,321 @@
...
+ * -------------------------------------------------------------------------------------------+
+ *
+ * -------+---------------------------+-------------+-------------+---------------------------+
+ *        |             |                           |             |                           |
+ * length | Children length           | Match idx   | Value       | Children                  |
+ *        |             |                           |             |                           |
+ * -------+---------------------------+-------------+-------------+---------------------------+
+ * </pre>
+ */
+public class MatchIndex {
```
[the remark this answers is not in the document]
rossbacher: Yes, I will add some documentation.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+import com.airbnb.deeplinkdispatch.base.MatchIndex
+import com.airbnb.deeplinkdispatch.base.MatchIndex.*
+import java.io.OutputStream
+import java.nio.charset.Charset
+import kotlin.text.Charsets.UTF_8
+
+data class UriMatch(val uri: DeepLinkUri, val matchId: Int)
+
+@kotlin.ExperimentalUnsignedTypes
+open class TrieNode(open val id: String = "", val type: UByte, open val placeholder: Boolean = false) {
```
elihart: can a null id be used instead of empty string? generally empty string is a red flag to me in kotlin, but it may makes sense for this implementation

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    val children = mutableSetOf<TrieNode>()
+    var match: UriMatch? = null
+        set(value) {
+            if (match != null) error("Ambiguous URI. Same match for two URIs ($match vs $value)") else field = value
```
elihart: use `field` to reference the current value instead of `match`

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+    fun addNode(node: TrieNode): TrieNode {
+        return if (children.add(node)) node else children.find { it.equals(node) }!!
```
elihart: `!!` is a bit of a red flag. can use `first` instead if that is the behavior you want

also, `==` instead of `.equals`

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+        for (childByteArray in childrenByteArrays) {
+            childByteArray.copyInto(resultByteArray, position)
+            position += childByteArray.size
+        }
+        return resultByteArray
+    }
+
+    private fun arrayLength(childArrays: List<UByteArray>, value: UByteArray, header: UByteArray): Int {
+        var length = header.size + value.size
+        childArrays.forEach { child ->
```
elihart:
```
val childrenLength = childArrays.sumBy { it.size }
return length + childrenLength

File: deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt [hunk 9]
```diff
@@ -0,0 +1,143 @@
...
+            length += child.size
+        }
+        return length
+    }
+
+    private fun generateChildrenByteArrays(): List<UByteArray> = children.map { it.toByteArray() }
+
+    private fun generateHeader(type: UByte, value: UByteArray, children: List<UByteArray>? = null, match: UriMatch?): UByteArray {
+        var childrenLength: UInt = 0u
+        children?.forEach { child ->
```
elihart: use `sumBy` here

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+            childrenLength += child.size.toUInt()
+        }
+        val header = UByteArray(HEADER_LENGTH)
```
elihart:
```
return UByteArray(HEADER_LENGTH).apply {
   set(0, type)
   ...
}

File: deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt [hunk 11]
```diff
@@ -0,0 +1,143 @@
...
+        header.writeUIntAt(2, childrenLength)
+        header.writeUShortAt(6, if (match == null) UShort.MAX_VALUE else match.matchId.toUShort())
+        return header
+    }
+
+}
+
+private val MAX_EXPOT_STRING_SIZE = 60000
+
+data class Root(val bla: String = "") : TrieNode(ROOT_VALUE, TYPE_ROOT.toUByte()) {
```
elihart: update `bla`?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+    fun writeToOutoutStream(openOutputStream: OutputStream) {
+        openOutputStream.write(this.toByteArray().toByteArray())
+    }
+
+    /**
+     * Convert the byte array into a string return as max 60k length sset of strings.
+     */
+    fun getStrings(): List<String> {
+        return String(bytes = this.toByteArray().toByteArray(), charset = Charset.forName(MatchIndex.MATCH_INDEX_ENCODING)).chunked(MAX_EXPOT_STRING_SIZE)
```
elihart: `toByteArray().toByteArray()` duplicate?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
```diff
@@ -0,0 +1,143 @@
...
+                node = node.addNode(PathSegment(pathSegment, hasPlaceholders(pathSegment)))
+            }
+            node.match = UriMatch(deeplinkUri, matchIndex)
+        }
+    }
+
+    private fun hasPlaceholders(pathSegment: String): Boolean {
+        val placeholderStart = pathSegment.contains("{")
+        val placeholderEnd = pathSegment.contains("}")
+        require(!(placeholderStart xor placeholderEnd)) { "Placeholders need to be complete (need to contain { and }" }
```
elihart: no closing paren in error message

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -0,0 +1,321 @@
...
+  public Match matchUri(@NonNull List<UrlElement> elements, @Nullable Map<String, String>
+      placeholders, int elementIndex, int elementStartPosition, int parentBoundryPos) {
+    Match match = null;
+    int currentElementStartPosition = elementStartPosition;
+    do {
+      UrlElement urlElement = elements.get(elementIndex);
+      String compareResult = compareValue(currentElementStartPosition, urlElement.getType(),
+          urlElement.getValue());
+      if (compareResult != null) {
+        // Need to hand down a new Hashmap, but only if we already have one.
```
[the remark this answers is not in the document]
rossbacher: No, but the code around it changed so the comment does not make the most sense anymore, let me update it.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -20,24 +21,31 @@
...
 
   private DeepLinkEntry findEntry(String uriString) {
-    for (Parser loader : loaders) {
-      DeepLinkEntry entry = loader.parseUri(uriString);
-      if (entry != null) {
-        return entry;
+    SchemeHostAndPath schemeHostAndPath = new SchemeHostAndPath(DeepLinkUri.parse(uriString));
+    DeepLinkEntry entryRegExpMatch = null;
+    DeepLinkEntry entryIdxMatch = null;
+    long regExSearchStart = SystemClock.elapsedRealtime();
```
BenSchwab: Remove these time logs?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/SchemeHostAndPath.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.airbnb.deeplinkdispatch.base.MatchIndex
+
+/**
+ * Used to encapsulate the scheme host and path of a DeepLinkUri into a type.
+ *
+ * All operations are happening on the UI thread by definition so it is ok to make the lazies
+ * not thread safe for speed improvement.
+ */
+class SchemeHostAndPath(val uri: DeepLinkUri) {
+    val schemeHostAndPath by lazy(LazyThreadSafetyMode.NONE) {
```
BenSchwab: This seems unused in the library, is this intended as part of the public api for convenience?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.java
```diff
@@ -0,0 +1,49 @@
...
+      return new byte[]{};
+    }
+    if (strings.length == 1) {
+      try {
+        return strings[0].getBytes(MatchIndex.MATCH_INDEX_ENCODING);
+      } catch (UnsupportedEncodingException e) {
+        // Do nothing. Encoding is hardcoded and default.
+      }
+    }
+    StringBuilder fullString = new StringBuilder(strings[0].length() * 2);
```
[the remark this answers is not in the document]
rossbacher: Yep, that was old testing code, thx for finding this.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -20,24 +21,31 @@
...
+  public BaseDeepLinkDelegate(List<? extends BaseLoader> loaders) {
     this.loaders = loaders;
   }
 
   private DeepLinkEntry findEntry(String uriString) {
-    for (Parser loader : loaders) {
-      DeepLinkEntry entry = loader.parseUri(uriString);
-      if (entry != null) {
-        return entry;
+    SchemeHostAndPath schemeHostAndPath = new SchemeHostAndPath(DeepLinkUri.parse(uriString));
```
BenSchwab: Looks like this is unused

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/SchemeHostAndPath.kt
```diff
@@ -0,0 +1,23 @@
...
+        uri.scheme() + "://" + uri.encodedHost() + uri.encodedPath()
+    }
+
+    val matchList by lazy(LazyThreadSafetyMode.NONE) {
```
BenSchwab: In the current code, this seems like it is always used whenever a `SchemeHostAndPath`, so might not need to be lazy.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
rossbacher: Leftover... :-)

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/UrlTrie.kt
rossbacher: No, the naming is a problem. I'll fix the naming

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/SchemeHostAndPath.kt
```diff
@@ -0,0 +1,23 @@
...
+        val list = mutableListOf(UrlElement(MatchIndex.TYPE_ROOT,MatchIndex.ROOT_VALUE.toByteArray()),UrlElement(MatchIndex.TYPE_SCHEME, uri.scheme().toByteArray()),UrlElement(MatchIndex.TYPE_HOST, uri.encodedHost().toByteArray()))
+        uri.encodedPathSegments().forEach({pathElement -> list.add(UrlElement(MatchIndex.TYPE_PATH_SEGMENT, pathElement.toByteArray()))})
+        list
+    }
+}
+
+data class UrlElement(val type: Byte, val value: ByteArray)
```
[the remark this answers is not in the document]
rossbacher: No, changed.

deeplinkdispatch/src/test/java/com/airbnb/deeplinkdispatch/DeepLinkEntryTest.java
```diff
@@ -10,136 +13,196 @@
 public class DeepLinkEntryTest {
```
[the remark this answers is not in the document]
rossbacher: I still need to write those. :-(

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/SchemeHostAndPath.kt
rossbacher: It was the original functionality of this class, I guess we don't need it anymore.

PR #269, opened 2020-01-03
Added benchmark tests using the AndroidX benchmark library

rossbacher:
- Moved ScaleTestActivity to it's own lib so it can be benchmarked (sample-benchmarkable-library)
- Added benchmark tests fow newing the registry, matching different URIs in the index and creating an result (sample-benchmark)
- Updated readme with baseline benchmark results on Pixel 2 running Android 10
- Updated compile SDK etc.
- Error message in DeeplinkResult is never null so made it non null
- createResult is never retuning null so made the return non null
- Renamed resources to avoid collisions


README.md
```diff
@@ -351,7 +376,7 @@ Note it is possible to call directly `adb shell am ...` however this seems to mi
 ## License
 
 ```
-Copyright 2015 Airbnb, Inc.
+Copyright 2015-2019 Airbnb, Inc.
```
BenSchwab: 2020? :)

sample-benchmark/benchmark-proguard-rules.pro
```diff
@@ -0,0 +1,37 @@
...
+
+-keepattributes *Annotation*
+
+-dontnote junit.framework.**
+-dontnote junit.runner.**
+
+-dontwarn androidx.test.**
+-dontwarn org.junit.**
+-dontwarn org.hamcrest.**
+-dontwarn com.squareup.javawriter.JavaWriter
```
BenSchwab: Shouldn't this only be on the annotation processor compile path? Why is proguard looking at it?

sample-benchmark/src/androidTest/java/com/airbnb/deeplinkdispatch/sample/benchmark/DeeplinkBenchmarks.kt
```diff
@@ -0,0 +1,120 @@
...
+/**
+* Benchmark, which will execute on an Android device.
+*
+* The body of [BenchmarkRule.measureRepeated] is measured in a loop, and Studio will
+* output the result. Modify your code to see how it affects performance.
+*/
+@RunWith(AndroidJUnit4::class)
+class DeeplinkBenchmarks {
+
+    private val DEEPLINK_1 = "dld://methodDeepLink1/test1234"
```
BenSchwab: nit: move to companion object and make `const val` (doesn't really matter since this is a test, but practical implication is that generates only a static field vs accessor function with backing field)

sample-benchmark/src/androidTest/java/com/airbnb/deeplinkdispatch/sample/benchmark/DeeplinkBenchmarks.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.airbnb.deeplinkdispatch.sample.benchmarkable.ScaleTestActivity
+import org.junit.Assert
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+* Benchmark, which will execute on an Android device.
+*
+* The body of [BenchmarkRule.measureRepeated] is measured in a loop, and Studio will
```
BenSchwab: Looks like some of this comment may be copied? Most of it seems useful besides the "modify your code to see how it affects performance"

sample-benchmark/benchmark-proguard-rules.pro
rossbacher: This is the default file generated by the new module wizard in AS. We are not using it anyway.

sample-benchmark/src/androidTest/java/com/airbnb/deeplinkdispatch/sample/benchmark/DeeplinkBenchmarks.kt
rossbacher: I think that was from the default test added by the wizard, let me clean that up.

PR #270, opened 2020-01-03
Rename `loader` -> `registry`

rossbacher:
Just renaming references and method names and comments, after we renamed the actual generated class.


README.md
```diff
@@ -205,19 +205,19 @@ dependencies {
...
 public class AppDeepLinkModule {
 }
 ```
 
-**Optional**: If your Android application contains multiple modules (eg. separated Android library projects), you'll want to add one `@DeepLinkModule` class for every module in your application, so that DeepLinkDispatch can collect all your annotations in one "loader" class per module:
+**Optional**: If your Android application contains multiple modules (eg. separated Android library projects), you'll want to add one `@DeepLinkModule` class for every module in your application, so that DeepLinkDispatch can collect all your annotations in one "Registry" class per module:
 
 ```java
-/** This will generate a LibraryDeepLinkModuleLoader class */
+/** This will generate a LibraryDeepLinkModuleLRegistry class */
```
BenSchwab:
```suggestion
/** This will generate a LibraryDeepLinkModuleRegistry class */
```

rossbacher: Thx for the fix.

PR #273, opened 2020-02-04
Fix Readme : rename registry to loader

hashlin:
Registry is already renamed to Loader.
I updated Readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,19 +207,19 @@ dependencies {
 
 _For **Kotlin** you should use_ `kapt` _instead of_ `annotationProcessor`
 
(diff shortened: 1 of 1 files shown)


rossbacher:
Actually no, in 4.0 we did rename the `Loader` to `Registry` I did search our code for `Loader` and could not find any (as it was all moved to `Registry` if you are still seeing `Loader` somewhere you might be using mixed dependencies e.g. an older version of the processor and a newer version of the lib.

The are *NOT* compatible and will not match correctly you must update them in unison.

rossbacher:
Actually in 5.x this was all renamed (which is in beta right now). Please use the beta and the documentation matches.

PR #274, opened 2020-02-13
Dynamic Path Segment Matching

adellhk:
Given: 

* Inject a `mapOf("some-key" to "bar", "high-key" to "hyphy")` [into BaseDeepLinkDelegate](https://github.com/airbnb/DeepLinkDispatch/pull/274/files#diff-79d4cabc99e211f7dc669529d9037aebR52).
* Annotate a class or Intent-returning method with `@DeepLink("https://www.example.com/foo/%%%some-key%%%/baz")` [example](https://github.com/airbnb/DeepLinkDispatch/pull/274/files#diff-ff82cdd41b0c2c12f5b1917b54fbd0a0R36)
* Annotate another class or Intent-returning method with `@DeepLink("https://www.example.com/you/only/%%%some-key%%%/%%%high-key%%%")`

Then:
* `baseDeepLinkDelegate.supporstUri("https://www.example.com/foo/bar/baz") //true`
* `baseDeepLinkDelegate.supportsUri("https://www.example.com/foo/hyphy/baz") //false`
* `baseDeepLinkDelegate.supportsUri("https://www.example.com/you/only/bar/hyphy") //true`

The motivation here is that users should be able to declare DeepLinks with a [replaceable path segment](https://github.com/airbnb/DeepLinkDispatch/pull/274/files#diff-b337e74e7e95cc7604eda7405470a206R74) (or segments) and then replace them at runtime (e.g. based on build type or a configuration flag).

[This test](https://github.com/airbnb/DeepLinkDispatch/pull/274/files#diff-cf1d85a2710e39c2b1b2ad89a95d4abfR176) covers the happy path of this change (other tests cover unhappy paths):
```java
 public void testPlaceholderSubstitution() {
    Map<String, String> pathVariableReplacements = new HashMap<>();
    pathVariableReplacements.put("replaceable-path-variable", "obamaOs");
    DeepLinkDelegate deepLinkDelegate = new DeepLinkDelegate(new SampleModuleRegistry(),
      new LibraryDeepLinkModuleRegistry(), new BenchmarkDeepLinkModuleRegistry(), pathVariableReplacements);
    assertThat(deepLinkDelegate.supportsUri("https://www.example.com/capnMcCains/bar"), equalTo(false));
    assertThat(deepLinkDelegate.supportsUri("https://www.example.com/obamaOs/bar"), equalTo(true));
  }
```

`./gradlew check test`:
Screen Shot 2020-02-12 at 9 24 46 PM


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
```diff
@@ -87,6 +87,9 @@
   /** Canonical URL. */
   private final String url;
 
+  /** Cached copy of the last Builder, set in {@link #parse }. */
+  private static Builder lastBuilder;
```
BenSchwab: Can we delete this now?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -210,4 +236,47 @@ private static void notifyListener(Context context, boolean isError, Uri uri,
...
+  /**
+   * Ensure that every key-to-be-replaced declared by all registries have a corresponding key in
+   * the user's injected mapping of pathVariableReplacements. If not, throw an exception and tell
+   * the user which keys aren't present.
+   * @param registries
+   * @param pathVariableReplacements
+   */
+  private void validatePathVariableReplacements(List<? extends BaseRegistry> registries, Map<String,
+    String> pathVariableReplacements) {
+    HashSet<String> keysUnion = new HashSet<>();
```
BenSchwab: Is this pretty fast? We could do it in a background thead if slow.

Also I feel like we could write this a lot more clearly in kotlin. Seems like it could just be an internal util instead of a member of this class, and then write that in Kotlin?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
```diff
@@ -52,27 +57,70 @@ open class TreeNode(open val id: String, val type: UByte, open val placeholder:
...
+        PathSegmentReplacement(8u)
+    }
+
+    /**
+     * Transformation types:
+     * @return an ASCII encoding as a character for the flag for whichever transformation type
+     * applies.
+     */
+    private fun String.transformationType(): UByte = when {
+        startsWith("%%%") && endsWith("%%%") -> {
```
BenSchwab: Curious why `%%%` over something simpler like `<` `>`

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/BaseRegistry.kt
```diff
@@ -12,31 +12,30 @@ import com.airbnb.deeplinkdispatch.base.MatchIndex
...
 
     /**
      * Use a binary match index to match the given URL. Defaults can be find in [MatchIndex].
      *
      * @param deepLinkUri The input [DeepLinkUri] to match
      * @return A [DeepLinkEntry] if a match was found or null if not.
      */
-    fun idxMatch(deepLinkUri: DeepLinkUri?): DeepLinkEntry? {
+    @JvmOverloads
+    fun idxMatch(deepLinkUri: DeepLinkUri?, pathSegmentReplacements: Map<String, String> = mapOf()): DeepLinkEntry? {
```
BenSchwab: add kdoc param for `pathSegmentReplacements`?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
```diff
@@ -52,27 +57,70 @@ open class TreeNode(open val id: String, val type: UByte, open val placeholder:
...
+            TransformationType.PathSegmentReplacement.flag
+        }
+        contains("{") && contains("}") -> {
```
BenSchwab: Can a single path contain two of these replacements?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -12,41 +13,28 @@
 
 /**
  * This is a wrapper class around the byte array match index.
+ * <p>Byte array format is:</p>
+ * <table border="1">
```
BenSchwab: Does the table render nicer in IntelliJ? -- It's a lot harder to read raw.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
adellhk: oops

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/BaseRegistry.kt
adellhk: yes thanks!

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
adellhk: Yes, I will add another unit test!

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
adellhk: no, it's terrible everywhere. It will render in an actual javadoc. I think the old format (ascii art) is unmaintainable though.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
adellhk: I'm imagining a pretty small number of keys to union (like 1 per DLD instance), but we can totally do it on a background, esp since it will run at every DLD init.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
adellhk: I want it to be really obvious that these are non-standard path segments. It might be better to have differentiated opening/closing characters, though, such as `<<` and `>>`.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/BaseRegistry.kt
BenSchwab: Bump

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -128,83 +108,109 @@ public Match matchUri(@NonNull List<UrlElement> elements, @Nullable Map<String,
...
-  String compareValue(int elementStartPos, byte type, @NonNull byte[] value) {
+  @Nullable
+  private String compareValue(int elementStartPos, byte inboundUriComponentType, @NonNull byte[]
+    inboundValue, Map<String, String> pathSegmentReplacements) {
     // Placeholder always matches
-    int valuePos = elementStartPos + HEADER_LENGTH;
-    boolean containsPlaceholder = false;
+    int valueStartPos = elementStartPos + HEADER_LENGTH;
+    byte nodeMetadata = byteArray[elementStartPos];
+    boolean isComponentParam = isComponentParam(nodeMetadata);
```
BenSchwab: 👍 Thanks for the self documenting variables

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -128,83 +108,109 @@ public Match matchUri(@NonNull List<UrlElement> elements, @Nullable Map<String,
...
+        //Per com.airbnb.deeplinkdispatch.DeepLinkEntryTest.testEmptyPathPresentParams
+        //We expect an empty path to not be a match
+        inboundValue.length == 0
+      )) {
+        return null;
+      }
+      // i index over inboundValue array forward
+      // j index over search byte arrays inboundValue element forward
+      // k index over inboundValue array backward
+      for (int i = 0; i < inboundValue.length; i++) {
```
BenSchwab: nit: This function is getting long, might be nice to break into two smaller functions, one for  comparing componentParam, and one for configurablePathSegment

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
```diff
@@ -20,4 +24,42 @@ object Utils {
     private fun getBytes(inputStream: InputStream): ByteArray {
         return inputStream.readBytes()
     }
+
+    @JvmStatic
+    fun isComponentTypeRoot(nodeMetadata: Byte): Boolean =
```
BenSchwab: Would these be better colocated with the definition of the enum, instead of in a utils class?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
BenSchwab: Bump on considering:
1) background thread
2) moving to kotlin extension
3) instead of (1) base delegate could take a debug flag to enable / disable this validation

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -344,61 +350,89 @@ public int compare(DeepLinkAnnotatedElement element1, DeepLinkAnnotatedElement e
...
+        element.getMethod());
+
+      //Keep track of pathVariables added in a module so that we can check at runtime to ensure
+      //that all pathVariables have a corresponding entry provided to BaseDeepLinkDelegate.
+      for (String pathSegment : deeplinkUri.pathSegments()) {
+        if (pathSegment.startsWith(pathSegmentEnclosingSequence)
+          && pathSegment.endsWith(pathSegmentEnclosingSequence)) {
+          pathVariableKeys.add(pathSegment.substring(pathSegmentEnclosingSequence.length(),
+            pathSegment.length() - pathSegmentEnclosingSequence.length()));
+        } else if (pathSegment.contains(pathSegmentEnclosingSequence)) {
```
BenSchwab: Any other validations that would be good to add here?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
```diff
@@ -52,27 +55,77 @@ open class TreeNode(open val id: String, val type: UByte, open val placeholder:
         return resultByteArray
     }
 
+    /**
+     * Use a UByte for [flag] so that we can easily specify and read different
+     * combinations/permutations of flags with a bit mask.
+     */
+    enum class NodeMetadata(val flag: UByte) {
```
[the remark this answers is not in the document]
adellhk: I think @rossbacher originally intended ubiquitous UByte usage to get an extra bit of space per byte.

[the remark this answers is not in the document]
BenSchwab: imo since we don't need the extra bit yet, it would be better to not use an experimental api. We can also switch to that down the road.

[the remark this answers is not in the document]
adellhk: We also use `UByte`s throughout the new (in 2020, since I added Kotlin support everywhere in DLD) Kotlinized parts of DLD, so I think the conversation here expands to the rest of the library. Slacked you and Rossi about this.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -183,7 +210,7 @@ public DeepLinkResult dispatchFrom(Activity activity, Intent sourceIntent) {
       }
       return new DeepLinkResult(true, uriString, "", newIntent, taskStackBuilder, deepLinkEntry);
     } catch (NoSuchMethodException exception) {
-      return new DeepLinkResult(false, uriString, "Deep link to non-existent method: "
+      return new DeepLinkResult(false, uriString, "Dee3p link to non-existent method: "
```
rossbacher: I guess that is a typo?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -23,25 +23,51 @@
...
+    configurablePathSegmentReplacements = new HashMap<>();
+  }
+
+  public BaseDeepLinkDelegate(
+    List<? extends BaseRegistry> registries,
+    Map<String, String> configurablePathSegmentReplacements
+  ) {
+    this.registries = registries;
+    this.configurablePathSegmentReplacements = configurablePathSegmentReplacements;
+    ValidationUtilsKt.validateConfigurablePathSegmentReplacements(registries,
```
rossbacher: I kinda don't want to do that on every start, this is something that cannot change after we build the app, isn't there a way to do this at compile time?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/BaseRegistry.kt
```diff
@@ -9,34 +9,40 @@ import com.airbnb.deeplinkdispatch.base.MatchIndex
...
 
     /**
      * Use a binary match index to match the given URL. Defaults can be find in [MatchIndex].
      *
-     * @param deepLinkUri The input [DeepLinkUri] to match
+     * @param deepLinkUri The inbound [DeepLinkUri] to match. In the typical use-case, this inbound
+     * URI comes from an Android Intent.
+     * @param pathSegmentReplacements originally injected via [BaseDeepLinkDelegate]. User provided
+     * replacement values for replaceable path segments. Ex: mapOf("replaceable" to "swish") and
+     * a://host/<replaceable> will result in a://host/swish being a match (and
```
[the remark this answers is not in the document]
adellhk: I believe this example is accurate

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
```diff
@@ -23,17 +25,18 @@ open class TreeNode(open val id: String, val type: UByte, open val placeholder:
...
-     * 1                                                    value length
-     * 2..5                                                children length
-     * 6..8                                                 match id
-     * 8..(8+value length)                                  value
-     * (8+value length)..((8+value length)+children length) children
+     * 0                                                    [NodeMetadata] flags; 1 byte
+     * 1                                                    length of value sub-array
+     * 2...5                                                length of node's children sub-array
+     * 6...7                                                match id
+     * 8..(8+value length)                                  actual value sub-array
```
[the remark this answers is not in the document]
adellhk: I mean sub-array of the entire byte array

PR #292, opened 2020-06-18
Fix issue with chunk length

rossbacher:
The Java/Kotlin compiler saves the string constant in a Modified UTF-8 byte array which has to be 16 bit addressable.
This PR adds a function that will chunk any given string based on it's Modified UTF-8 byte array representation.

This will allow us to have as few chunks as possible and also not accidentally creating a chrunk that is larger than the max size allowed in code.

- Added new function `chunkOnModifiedUtf8ByteSize` to chunk a string based on how long it's representation as Modified UTF-8 byte array would be, as this is what the java compiler uses to store the string in code and what the variable size limitation in based upon.
- Added tests for new function
- Update `gradleMavenPublishPlugin` and `agp` to make CI pass.


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
```diff
@@ -6,6 +6,41 @@ import com.airbnb.deeplinkdispatch.configurablePathSegmentSuffix
...
 
+/**
+ * Chunk a CharSequence based on how long it's Modified UTF-8
+ * (https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8) ByteArray representaiton would be.
+ */
+fun CharSequence.chunkOnModifiedUtf8ByteSize(chunkSize: Int): List<CharSequence> {
+    require (chunkSize >= 3) {
+        "UTF-8 chars can be up to 3 bytes wide. Minumum chunk size is 3 bytes."
+    }
+    var modifiedUtf8BlockSize = 0
```
adellhk, line 17: I think this and the variable on L25 should be named consistently to each other so they're easier to associate, potentially `next(Chunk|Block)ModifiedUtf8Size`

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
```diff
@@ -6,6 +6,41 @@ import com.airbnb.deeplinkdispatch.configurablePathSegmentSuffix
...
+    val result = mutableListOf<CharSequence>()
+    var blockStart = 0
+    var i = 0
+    while (i in 0 until this.length) {
+        // Get the byte array for every character and check how many bytes this would take up.
+        // U+0000 is encoded as two bytes C080 in mModified UTF-8, which is used by Java to
+        // store strings in the string table in class files.
+        val charByteArraySize = this.substring(i, i + 1).toByteArray().size + if (this[i] == '\u0000') 1 else 0
```
adellhk, line 25: potentially `charModifiedUtf8Size`

deeplinkdispatch-base/src/test/java/com/airbnb/deeplinkdispatch/CharSequenceUtf8Test.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test(expected = IllegalArgumentException::class) fun testChunkToSmall(){
```
adellhk, line 12:
```suggestion
    @Test(expected = IllegalArgumentException::class) fun testChunkTooSmall(){
```

deeplinkdispatch-base/src/test/java/com/airbnb/deeplinkdispatch/CharSequenceUtf8Test.kt
```diff
@@ -0,0 +1,94 @@
...
+        val source = "1€\u0000"
+        val testResult = source.chunkOnModifiedUtf8ByteSize(5)
+        assertEquals(testResult.size, 2)
+        assertEquals(testResult[0].length, 2)
+        assertEquals(testResult[1].length, 1)
+        assertTrue(compareSourceAndChunked(source,testResult))
+    }
+
+    @Test fun testNormalAscii(){
+        val source = "0123456789"
```
adellhk, line 68: clever

deeplinkdispatch-base/src/test/java/com/airbnb/deeplinkdispatch/CharSequenceUtf8Test.kt
```diff
@@ -0,0 +1,94 @@
...
+        assertEquals(testResult.size, 1)
+        assertEquals(testResult[0].length, 1)
+        assertTrue(compareSourceAndChunked(source,testResult))
+    }
+
+    @Test fun testChunk(){
+        val source = "\u0000\u0000\u0000"
+        val testResult = source.chunkOnModifiedUtf8ByteSize(5)
+        assertEquals(testResult.size, 2)
+        assertEquals(testResult[0].length, 2)
```
allenchen1154, line 28: Would it make more sense to assert that each chunk's byte size is <= the argument provided to `chunkOnModifiedUtf8ByteSize()`? In theory that's what we care about, not the number of characters in each chunk.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
```diff
@@ -6,6 +6,41 @@ import com.airbnb.deeplinkdispatch.configurablePathSegmentSuffix
...
+ * (https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8) ByteArray representaiton would be.
+ */
+fun CharSequence.chunkOnModifiedUtf8ByteSize(chunkSize: Int): List<CharSequence> {
+    require (chunkSize >= 3) {
+        "UTF-8 chars can be up to 3 bytes wide. Minumum chunk size is 3 bytes."
+    }
+    var modifiedUtf8BlockSize = 0
+    val result = mutableListOf<CharSequence>()
+    var blockStart = 0
+    var i = 0
```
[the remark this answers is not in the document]
rossbacher: It can be. Good catch. I actually find this whole loop rather ugly and reconfigured it a few times to get it nicer and this was a left over from that.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
rossbacher: Good catch about the terminology, i did not even notice I used two.

deeplinkdispatch-base/src/test/java/com/airbnb/deeplinkdispatch/CharSequenceUtf8Test.kt
rossbacher: I was thinking about that but did not want to implement the chunk method in reverse to test the result but I think I do not have to do that, will add some.

deeplinkdispatch-base/src/test/java/com/airbnb/deeplinkdispatch/CharSequenceUtf8Test.kt
```diff
@@ -49,46 +54,57 @@ class CharSequenceUtf8Test {
...
+        assertTrue(testResult.checkModifiedUtf8ByteArraySize(100))
     }
 
     fun compareSourceAndChunked(source: CharSequence, chunked : List<CharSequence>) : Boolean {
         return source == chunked.joinToString(separator = "")
     }
+
+    fun List<CharSequence>.checkModifiedUtf8ByteArraySize(maxByteArraySize: Int): Boolean{
+        return all { charSequence ->
+            (charSequence.toString().toByteArray().size) + charSequence.count { char -> char == '\u0000' } <= maxByteArraySize}
```
allenchen1154, line 108: 👍

PR #293, opened 2020-06-19
Add the originating Elements to all generate source files.

rossbacher:
- Add the originating Elements to all generate source files.
- Update to non snapshot version of gradle-maven-publish-plugin as the snapshot version was removed.


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -517,6 +525,7 @@ public ParameterSpec apply(TypeElement typeElement) {
       .superclass(CLASS_BASE_DEEP_LINK_DELEGATE)
       .addMethod(constructor)
       .addMethod(constructorWithPathVariables)
+      .addOriginatingElement(originatingElement)
```
BenSchwab, line 528: I don't thin it matters since the registry classes are generated.. but it might be most accurate to mark them as originating elements as well.

PR #294, opened 2020-06-26
Match deterministically across registries

rossbacher:
Changed code so that all registries (there is one registry per module) will be checked for matches and the same rule (`concrete > placeholder > configurable path segment` rule is applied for all matches.

This is accomplished by sorting the list of matched url templates by "concreteness" (based on where it has placeholders/configurable path segments) The same way this is done while matching inside an index

- Added tests for DeeplinkEntry "concreteness" comparator
- Added tests for (other than before) matching more concrete match in Library (that is matched after the sample app gets matched)

Require the mapping for configurable path segments given to the `DeeplinkDelegate` to not just match the configurable path segments when given but also when not given.

### DIFF
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
@@ -15,6 +15,8 @@
  */
 package com.airbnb.deeplinkdispatch
 
+import kotlin.math.min
+
 class DeepLinkEntry(val uriTemplate: String,
                     val type: Type,
                     /**
@@ -29,6 +31,24 @@ class DeepLinkEntry(val uriTemplate: String,
 
     private val parameterMap: MutableMap<DeepLinkUri, Map<String, String>> = mutableMapOf()
 
+    private val firstConfigurablePathSegmentIndex: Int by lazy { uriTemplate.indexOf(configurablePathSegmentPrefixChar) }
+    private val firstPlaceholderIndex: Int by lazy { uriTemplate.indexOf(componentParamPrefixChar) }
+    private val firstNonConcreteIndex: Int by lazy {
+        if (firstPlaceholderIndex == -1 && firstConfigurablePathSegmentIndex == -1) {
+            -1
+        } else {
+            if (firstConfigurablePathSegmentIndex == -1) {
+                firstPlaceholderIndex
+            } else {
+                if(firstPlaceholderIndex == -1){
+                    firstConfigurablePathSegmentIndex
+                } else {
+                    min(firstConfigurablePathSegmentIndex, firstPlaceholderIndex)
+                }
+            }
+        }
+    }
+
     /**
      * Generates a map of parameters and the values from the given deep link.
      *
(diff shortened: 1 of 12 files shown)


deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -47,6 +50,8 @@
   public BaseDeepLinkDelegate(List<? extends BaseRegistry> registries) {
     this.registries = registries;
     configurablePathSegmentReplacements = new HashMap<>();
+    ValidationUtilsKt.validateConfigurablePathSegmentReplacements(registries,
```
rossbacher, line 53: @adellhk was this an oversight or did we leave that out on purpose?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -29,17 +31,61 @@ class DeepLinkEntry(val uriTemplate: String,
...
+    }
+
+    /**
+     * Whatever template has the first placeholder (and then configurable path segment) is the less
+     * concrete one.
+     * Because if they would have been all in the same index those elements would have been on the
+     * same level and in the same "list" of elements we compare in order.
+     * In this case the one with the more concete element would have won and the same is true here.
+     */
+    fun moreConcreteThan(compare: DeepLinkEntry): Int {
```
BenSchwab, line 77: implement `Comparable` instead?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -29,17 +31,61 @@ class DeepLinkEntry(val uriTemplate: String,
...
+    private val firstConfigurablePathSegmentIndex: Int by lazy { uriTemplate.indexOf(configurablePathSegmentPrefixChar) }
+    private val firstPlaceholderIndex: Int by lazy { uriTemplate.indexOf(componentParamPrefixChar) }
+    private val firstNonConcreteIndex: Int by lazy {
+        if (firstPlaceholderIndex == -1 && firstConfigurablePathSegmentIndex == -1) {
+            -1
+        } else {
+            if (firstConfigurablePathSegmentIndex == -1) {
+                firstPlaceholderIndex
+            } else {
+                if(firstPlaceholderIndex == -1){
```
BenSchwab, line 43:
```suggestion
                if (firstPlaceholderIndex == -1){
```

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -210,15 +215,34 @@ public DeepLinkResult dispatchFrom(Activity activity, Intent sourceIntent) {
...
+    // Found multiple matches. Sort matches by concreteness:
+    // No variable element > containing placeholders >  are a configurable path segment
+    Collections.sort(entryIdxMatches, new Comparator<DeepLinkEntry>() {
+      @Override
+      public int compare(DeepLinkEntry a, DeepLinkEntry b) {
+        return a.moreConcreteThan(b);
       }
+    });
+    if (entryIdxMatches.get(0).moreConcreteThan(entryIdxMatches.get(1)) == 0) {
+      Log.w(TAG, "More than one match with the same concreteness!! ("
```
BenSchwab, line 242: I guess this is a bit indeterministic which gets called. Would be nice to be able to hook into this warning log and crash in development (maybe could add optional constructor injected logger interface)

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
rossbacher: Yeah, I had that, and then I refactored it way. :-) I agree.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
rossbacher: Yeah, let me add that.

rossbacher: @BenSchwab I added that and a test for it. Let me know if that is what you ad in mind?

rossbacher:
I agree with the better duplicate handling during build, but that should be a different task as there are several issue with that. One of them being that right now it cannot differentiate between links with differently named placeholders that duplicate otherwise. e.g. `http://example.com/path1/{placeholder1}` and `http://example.com/path1/{placeholder2}`

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.java
```diff
@@ -25,6 +27,8 @@
   protected static final String TAG = "DeepLinkDelegate";
 
   protected final List<? extends BaseRegistry> registries;
+
+  protected final ErrorHandler errorHandler;
```
BenSchwab, line 31: If we use`@Nullable` annotations in the project, we can add one here.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/ErrorHandler.kt
```diff
@@ -0,0 +1,5 @@
+package com.airbnb.deeplinkdispatch
+
+interface ErrorHandler {
+    fun duplicateMatch(duplicatedMatches: List<DeepLinkEntry>)
```
BenSchwab, line 4: Might be helpful for debugging include the source uri?

Issue #303, opened 2020-12-04
Not supported valid URIs for deeplinks with same path and different number of parameters


cvillaseca:
Hi,

Since we updated the  library to version 5.X some deeplinks like `myapp://vehicle/{vehicleId}` have stopped working. The problem is that with the new version it doesn’t handle multiple links with the same structure, it only handles the first one is created in the registry class. For example:

``` java
public final class MyDeepLinkModuleRegistry extends BaseRegistry {
  public MyDeepLinkModuleRegistry() {
    super(Collections.unmodifiableList(Arrays.<DeepLinkEntry>asList(
      new DeepLinkEntry("myapp://vehicle/{make}/{model}", DeepLinkEntry.Type.METHOD, VehicleDetailNavigation.class, "newIntentForTaskStackBuilderMethods"),
      new DeepLinkEntry("myapp://vehicle/{vehicleId}", DeepLinkEntry.Type.METHOD, VehicleDetailNavigation.class, "newIntentForTaskStackBuilderMethods")
    )), Utils.readMatchIndexFromStrings( new String[] {matchIndex0(), }),
    new HashSet<String>(Arrays.<String>asList()));
  }
}
```

This is considered a valid deeplink

```
myapp://vehicle/vw/golf
```

But this deeplink won’t be valid. You can check this by using `DeepLinkDelegate.supportsUri("myapp://vehicle/123")`

```
myapp://vehicle/123
```

This was working in version 4.X but with version 5.X it's not.

Thank you 🙏

rossbacher:
I'll have a look at that. I don't think we have a test case for this and thus it might have been missed.

stari4ek:
There is a chance that incremental build is involved.
I had same issue. Bunch of "foo://navigate/*" declared, but only one was added to registry.
Forcing rebuild the project solved the issue

rossbacher:
Sorry it took that long but this is fixed now and in release https://github.com/airbnb/DeepLinkDispatch/releases/tag/5.3.0.

cvillaseca:
Thank you for the fix @rossbacher 🙏

PR #308, opened 2021-03-18
Fix placeholder match

rossbacher:
Because of the way the index is sorted two deeplinks that are only differentiated by the number of placeholders in the match like:

`app://host/{a}{b}`and
`app://host/{c}`

would have the shorter one not match if the placeholder in question is lexicographic after the ones in the longer one.

This e.g. would have worked:

`app://host/{c}{b}`and
`app://host/{a}`

This fixes this by checking all elements on on one level of the trie before reporting no match.

This fixes https://github.com/airbnb/DeepLinkDispatch/issues/303

### DIFF
--- a/deeplinkdispatch/src/test/java/com/airbnb/deeplinkdispatch/DeepLinkEntryTest.java
+++ b/deeplinkdispatch/src/test/java/com/airbnb/deeplinkdispatch/DeepLinkEntryTest.java
@@ -31,7 +31,7 @@ public class DeepLinkEntryTest {
 
   /**
    * This test makes sure that both version of the parametrized strings are matched correctly.
-   * As mach groups are sorted alphabetically in the index we need to make sure that the
+   * As match groups are sorted alphabetically in the index we need to make sure that the
    * shorter one is in the index behind the longer one (as otherwise the test would match
    * correctly anyway)
    */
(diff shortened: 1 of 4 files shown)


dependencies.gradle
```diff
@@ -2,16 +2,16 @@ def versions = [
...
     appCompatVersion             : '1.1.0',
     localBroadcastManagerVersion : '1.0.0',
-    roboelectricVersion          : '4.3.1',
+    roboelectricVersion          : '4.5.1',
     benchmarkVersion             : '1.0.0'
 ]
 
 ext.versions = versions
 ext.androidConfig = [
-    agpVersion                          : '4.0.0',
```
elihart, line 11: thanks for bumping these too

deeplinkdispatch/src/test/java/com/airbnb/deeplinkdispatch/DeepLinkEntryTest.java
```diff
@@ -29,6 +29,33 @@
     assertThat(parameters.get("param2")).isEqualTo("alice");
   }
 
+  /**
+   * This test makes sure that both version of the parametrized strings are matched correctly.
+   * As mach groups are sorted alphabetically in the index we need to make sure that the
+   * shorter one is in the index behind the longer one (as otherwise the test would match
+   * correctly anyway)
+   */
+  @Test public void testOneAndTwoParamsSubPath() {
```
elihart, line 38: 🎉

elihart (approved):
thanks for fixing!

PR #310, opened 2021-04-07
Updated README.md to fix a small typo

nateshmbhat:


rossbacher:
this is not a typo, the query parameters do not need to be part of the template given in the annotation.

e.g. see this test: https://github.com/airbnb/DeepLinkDispatch/blob/master/sample/src/test/java/com/airbnb/deeplinkdispatch/sample/MainActivityTest.java#L94

PR #312, opened 2021-05-12
Embed match info into index

rossbacher:
Embed match information into the match index. This way we do not need to codegen all the `DeepLinkEntry` constructions into the generated `DeepLinkRegistry` classes.

This change will create the `DeepLinkEntry` object on the fly with data that is stored inside the index (vs just an id stored in the index that was used to look up the `DeepLinkEntry`).

For this the `matchId` entry was removed from the match index and a new entry which contains the deeplink url template, the class the deeplink is registered in as well as the method to call on that class (if not a class level deeplink) was added.

This leads to 5x speed improvement in index initialization (in our example with 2000 deeplinks index init went from ~12ms down to ~2 ms.

This will also remove the upper boundary for handled deeplinks from DLD as the number of generated `DeepLinkEntry` constructor calls can lead to the Registry classes not being compilable anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,24 +371,23 @@ kapt {
 
 Starting with v5 DeeplinkDispatch is designed to be very fast in resolving deep links even if there are a lot of them. To ensure we do not regress from this benchmark tests using [`androidx.benchmark`](https://developer.android.com/studio/profile/benchmark#top_of_page) were added.
 
-It is testing the `ScaleTestActivity` in the `sample-benchmarkable-library` which has 2000 deep links. For those on a Pixel 2 running Android 10 we expect the following results:
+It is testing the `ScaleTestActivity` in the `sample-benchmarkable-library` which has 2000 deep links. For those on a Pixel 2 running Android 11 we expect the following results:
 
 ```text
 Started running tests
-Timed out waiting for process to appear on google-pixel_2-FA7AW1A04466.
-benchmark:        11,520 ns DeeplinkBenchmarks.match1
-benchmark:       241,406 ns DeeplinkBenchmarks.match500
-benchmark:    12,067,970 ns DeeplinkBenchmarks.newRegistry
-benchmark:        12,076 ns DeeplinkBenchmarks.match1000
-benchmark:       140,000 ns DeeplinkBenchmarks.match1500
-benchmark:       273,230 ns DeeplinkBenchmarks.match2000
-benchmark:       148,750 ns DeeplinkBenchmarks.createResultDeeplink1
-benchmark:        11,375 ns DeeplinkBenchmarks.parseDeeplinkUrl
 
+benchmark:        11,716 ns DeeplinkBenchmarks.match1
+benchmark:       139,375 ns DeeplinkBenchmarks.match500
+benchmark:     2,163,907 ns DeeplinkBenchmarks.newRegistry
+benchmark:        23,035 ns DeeplinkBenchmarks.match1000
+benchmark:       152,969 ns DeeplinkBenchmarks.match1500
+benchmark:       278,906 ns DeeplinkBenchmarks.match2000
+benchmark:       162,604 ns DeeplinkBenchmarks.createResultDeeplink1
+benchmark:        11,774 ns DeeplinkBenchmarks.parseDeeplinkUrl
 Tests ran to completion.
 ```
 
-As you can see it takes us about 11ms to create the registry with 2000 entries. A lookup can be done in sub 1ms usually and `createResult`, which includes the lookup for the `match1` case plus actually calling the method that was annotated, can be done in 0.2ms. 
+As you can see it takes us about 11ms to create the registry with 2000 entries. A lookup can be done in sub 1ms usually and `createResult`, which includes the lookup for the `match1` case plus actually calling the method that was annotated, can be done in under 0.2ms.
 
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -371,24 +371,23 @@ kapt {
...
+benchmark:        23,035 ns DeeplinkBenchmarks.match1000
+benchmark:       152,969 ns DeeplinkBenchmarks.match1500
+benchmark:       278,906 ns DeeplinkBenchmarks.match2000
+benchmark:       162,604 ns DeeplinkBenchmarks.createResultDeeplink1
+benchmark:        11,774 ns DeeplinkBenchmarks.parseDeeplinkUrl
 Tests ran to completion.
 ```
 
-As you can see it takes us about 11ms to create the registry with 2000 entries. A lookup can be done in sub 1ms usually and `createResult`, which includes the lookup for the `match1` case plus actually calling the method that was annotated, can be done in 0.2ms. 
+As you can see it takes us about 11ms to create the registry with 2000 entries. A lookup can be done in sub 1ms usually and `createResult`, which includes the lookup for the `match1` case plus actually calling the method that was annotated, can be done in under 0.2ms.
```
elihart, line 390:
```suggestion
As you can see it takes us about 2ms to create the registry with 2000 entries. A lookup can be done in sub 1ms usually and `createResult`, which includes the lookup for the `match1` case plus actually calling the method that was annotated, can be done in under 0.2ms.
```

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/UrlTree.kt
```diff
@@ -128,6 +140,50 @@ data class Host(override val id: String) : TreeNode(id = id, metadata = NodeMeta
...
+ * (2+url template length)+2                                       classname length
+ * (2+url template length)+2..
+ * (2+url template length)+2+classname length                      class name
+ * (2+url template length)+2+classname length..
+ * (2+url template length)+2+classname length+1                    method name length (can be 0)
+ * (2+url template length)+2+classname length+1..
+ * (2+url template length)+2+classname length+1+method name length method name
+ */
+fun matchByteArray(match: UriMatch?): UByteArray {
+    return match?.let {
```
elihart, line 157: avoid the nesting and using `it` in such  large block
```suggestion
    if (match == null) return UByteArray(0)
```

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -308,7 +354,10 @@ private int getNextElementStartPosition(int elementStartPos, int parentBoundaryP
    * @return The first elementStartPos that is not part of the parent element anymore.
    */
   private int getElementBoundaryPos(int elementStartPos) {
-    return elementStartPos + HEADER_LENGTH + getValueLength(elementStartPos)
+    return elementStartPos
```
elihart, line 357: would be nice to dedupe and chain this like `return getChildrenPos(elementStartPos) + getChildrenLength(elementStartPos)`

dependencies.gradle
```diff
@@ -1,43 +1,43 @@
 def versions = [
     kotlinVersion                : '1.3.72',
```
elihart, line 2: this kotlin version is fairly out of date, the newer versions have made changes to unsigned number handling and I think it would be good to get those updates in as you do so much unsigned manipulation

dependencies.gradle
```diff
@@ -1,43 +1,43 @@
...
 ext.deps = [
-    autoCommon               : "com.google.auto:auto-common:0.10",
+    autoCommon               : "com.google.auto:auto-common:1.0",
     androidPlugin            : "com.android.tools.build:gradle:$androidConfig.agpVersion",
     appCompat                : "androidx.appcompat:appcompat:$versions.appCompatVersion",
     localBroadcastManager    : "androidx.localbroadcastmanager:localbroadcastmanager:$versions.localBroadcastManagerVersion",
     kotlinStdLib             : "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$versions.kotlinVersion",
     kotlinGradlePlugin       : "org.jetbrains.kotlin:kotlin-gradle-plugin:$versions.kotlinVersion",
-    javaPoet                 : 'com.squareup:javapoet:1.12.1',
+    javaPoet                 : 'com.squareup:javapoet:1.13.0',
```
elihart, line 24: thanks for updating dependencies

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-all.zip
```
elihart, line 6: why not 7.0.1?

elihart (approved):
Really great improvements, nice work

dependencies.gradle
rossbacher: I tried 1.5 and 1.4 but ran into issues that i did not want to deal with, I can give it another try.

gradle/wrapper/gradle-wrapper.properties
rossbacher: We can go there, i thought there might be some issues with 7, but I'll check it out.

PR #314, opened 2021-05-14
Use separate Match class, update Kotlin and OKIO

rossbacher:
This is updating some dependencies, which required converting tests to Kotlin.
It also create a new `DeepLinkMatchResult` class that is used instead of `DeepLinkEntry` to report the match result.
This separates the `DeepLinkEntry` from the url -> placeholder map that is part of the result and was previously added to the entry, which is just representing the deeplink as it exists in the index.


elihart (approved):
Great updates 🎉

PR #315, opened 2021-05-19
Enable placeholders in custom annotation prefixes.

rossbacher:
This change enables the usage of placeholders in custom annotation prefixes. 

Internally DLD is creating a separate deeplink entry for every deeplink multiplied by the number prefixes.

e.g. if you have 100 places which use your custom `WebDeepLink` which is defined like this:

```java
@DeepLinkSpec(prefix = { "http://example.com", "https://example.com" })
@Retention(RetentionPolicy.RUNTIME)
public @interface WebDeepLink {  String[] value();}
```

it will crate 200 entries as there are 2 prefixes per usage. If you have many prefixes e.g. for country domain differences this can add up and create a lot of entires that are very similar. 

DLD is capable of handling that many entires but it is still not very efficient.

With this change it will allow you to write something like this:

```java
@DeepLinkSpec(prefix = { "http{scheme}://example.com" })
@Retention(RetentionPolicy.RUNTIME)
public @interface WebDeepLink {  String[] value();}
```

This will match `http` and `https` urls (but also basically any url scheme that start with `http` and it will only create one entry in the match index.

If you also have country prefixes for the hosts you can write something like this:

```java
@DeepLinkSpec(prefix = { "http{scheme}://{host_prefix}.example.com" })
@Retention(RetentionPolicy.RUNTIME)
public @interface WebDeepLink {  String[] value();}
```

This will match hosts that are `.*example.com` without having to list them all and still with just one entry in the match index.

### DIFF
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
@@ -37,9 +37,12 @@ import java.util.Set;
 import okio.Buffer;
 
 /**
- * Adapted from OkHttp's HttpUrl class. Only change is to allow any scheme, instead of just http or
- * https.
- *https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/HttpUri.java
+ * Adapted from OkHttp's HttpUrl class. Changes are:
+ *  * Allow any scheme, instead of just http or https.
+ *  * when parsing via parseTemplate(url) allow placeholder chars({}) in url.
+ *
+ * https://github.com/square/okhttp/blob/master/okhttp/src/main/java/com/squareup/okhttp/
+ * HttpUri.java
  */
 public final class DeepLinkUri {
   private static final char[] HEX_DIGITS =
@@ -362,7 +365,7 @@ public final class DeepLinkUri {
   /** Returns the URL that would be retrieved by following {@code link} from this URL. */
   DeepLinkUri resolve(String link) {
     Builder builder = new Builder();
-    Builder.ParseResult result = builder.parse(this, link);
+    Builder.ParseResult result = builder.parse(this, link, false);
     return result == Builder.ParseResult.SUCCESS ? builder.build() : null;
   }
 
@@ -385,8 +388,22 @@ public final class DeepLinkUri {
    * URL, or null if it isn't.
    */
   public static DeepLinkUri parse(String url) {
+    return parse(url, false);
+  }
+
+  /**
+   * Like parse(String url) but allow '{' '}' chars in scheme to allow using placeholders in scheme.
+   *
+   * @param url
+   * @return
+   */
+  public static DeepLinkUri parseTemplate(String url) {
+    return parse(url, true);
+  }
+
+  private static DeepLinkUri parse(String url, boolean allowPlaceholderInSchem) {
     Builder builder = new Builder();
-    Builder.ParseResult result = builder.parse(null, url);
+    Builder.ParseResult result = builder.parse(null, url, allowPlaceholderInSchem);
     return result == Builder.ParseResult.SUCCESS ? builder.build() : null;
   }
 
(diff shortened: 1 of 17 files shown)


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
```diff
@@ -385,8 +388,22 @@ Builder newBuilder() {
...
+   * @return
+   */
+  public static DeepLinkUri parseTemplate(String url) {
+    return parse(url, true);
+  }
+
+  private static DeepLinkUri parse(String url, boolean allowPlaceholderInSchem) {
     Builder builder = new Builder();
-    Builder.ParseResult result = builder.parse(null, url);
+    Builder.ParseResult result = builder.parse(null, url, allowPlaceholderInSchem);
```
elihart, line 406:
```suggestion
    Builder.ParseResult result = builder.parse(null, url, allowPlaceholderInScheme);
```

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
```diff
@@ -287,46 +289,68 @@ private CompareResult compareConfigurablePathSegment(@NonNull byte[] inboundValu
...
+    // k index over valueToMatch array backward
+    for (int i = 0; i < valueToMatchLength; i++) {
+      // Edge case. All chars before placeholder match. And there are no more chars left in
+      // input string. But there are still chars left in value. In that case we need to look
+      // at the rest of the value as there might chars after the placeholder (e.g. pre{ph}post)
+      //or a placeholder that matches the empty string. e.g.
+      boolean fullyMatchedButThereAreMoreCharsInValue =
+        byteArray[valueStartPos + i] == valueToMatch[i]
+          && i == valueToMatchLength - 1
+          && valueLength > valueToMatchLength;
```
elihart, line 317: given how much you've been working in this class it would be great to convert it to kotlin

elihart (approved):
Cool improvement, glad to see this more efficient.

I just have questions from the design side - could it potentially be a security issue that this would allow bad schemes like `httpz` or other? It seems like it would be safer for the design to be more explicit in matching, like `prefix = { "[http]|[https]://airbnb.com", almost more like a regex. Or have a separate way to set valid options like `val validSchemes = "http,https"  ; prefix = { "{validSchemes}://airbnb.com"`

not sure how much of an issue that really is, but it came to mind.

My other question is if there are plans or possibilities to handle the tld with a placeholder too (which would require an allow list of placeholder values I guess)

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/MatchIndex.java
rossbacher: I did that originally and had performance problems making it about 60% slower. So I moved it back to java. I can eventually look at that again, but for now I think it is just not worth it, also given how low level that is.

I know back then we did think these were type conversions happening but I did not want to hunt them down yet. :-)

rossbacher:
@elihart I was thinking about those, but they actually require some bigger changes to the index and I did want to avoid that. I thought about it and I don't think they are required from a security point of view.

> I just have questions from the design side - could it potentially be a security issue that this would allow bad schemes like `httpz` or other? It seems like it would be safer for the design to be more explicit in matching, like `prefix = { "[http]|[https]://airbnb.com", almost more like a regex. Or have a separate way to set valid options like `val validSchemes = "http,https" ; prefix = { "{validSchemes}://airbnb.com"`

Actually the supported schemas are listed in the manifest still, so if I only list http and https I would not even get anything else to handle. But even if I did, it does not really matter. It would just mean that we handle more links then we want but the links are not "vetted" e.g. someone can already create a link with http://www.example.com/maliciousContent which would match `http://www.example.com/{placeholder}` and we need to (right now) be able to handle that. No harm is created even if we would handle `httpz://www.example.com/maliciousContent`
 
> not sure how much of an issue that really is, but it came to mind.
> 
> My other question is if there are plans or possibilities to handle the tld with a placeholder too (which would require an allow list of placeholder values I guess)

We can do that right now with those code. Just define it as `http{scheme}://www.example.{tld}` e.g. again we already have to handle all those URLs no harm can come from someone inventing a TLD and making us handle this.

Also again for TLDs we already have to filter via the manifest entry anyway.

PR #317, opened 2021-05-25
Fixes to template parsing

rossbacher:
- Allow support for characters that are not allowed in schemes when used in placeholders.
- Allow scheme to be just a placeholder or start with a placeholder 
- Add tests for all of these cases.


pmecho (approved):
Thanks for adding tests!

PR #321, opened 2021-06-04
Fix inner class matching (matching on methods in inner classes)

rossbacher:
- Add test to test inner class matching
- Remove nullability of activity class and throw exception (and crash) if class does not exist as this should never happen (unless it gets removed by Dexguard/Proguard/R8), this also removes this case from the error handler.
- Update Kotlin and AGP dependency to latest

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
@@ -348,7 +348,7 @@ public class DeepLinkProcessor extends AbstractProcessor {
       String uriTemplate = element.getUri();
 
       try {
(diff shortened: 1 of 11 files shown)


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.java
```diff
@@ -348,7 +348,7 @@ public int compare(DeepLinkAnnotatedElement element1, DeepLinkAnnotatedElement e
       String uriTemplate = element.getUri();
 
       try {
-        urisTrie.addToTrie(uriTemplate, activity.canonicalName(), element.getMethod());
+        urisTrie.addToTrie(uriTemplate, activity.reflectionName(), element.getMethod());
```
rossbacher, line 351: This is the actual fix.

BenSchwab (approved):

PR #326, opened 2021-06-30
Fix issue with annotated Kotlin function that were not inside a class when using ksp

rossbacher:
Make DeepLinkAnnotatedElement.annotatedClass an XMemberContainer as that also wraps the case when the method is not in an annotated class in the ksp case.

Added test for this usecase.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.kt
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.kt
@@ -16,6 +16,7 @@
 package com.airbnb.deeplinkdispatch
 
 import androidx.room.compiler.processing.XElement
+import androidx.room.compiler.processing.XMemberContainer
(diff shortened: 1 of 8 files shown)


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkAnnotatedElement.kt
```diff
@@ -16,16 +16,17 @@
...
+import androidx.room.compiler.processing.XMemberContainer
 import androidx.room.compiler.processing.XMethodElement
 import androidx.room.compiler.processing.XTypeElement
 import java.net.MalformedURLException
 
 sealed class DeepLinkAnnotatedElement @Throws(MalformedURLException::class) constructor(
     val uri: String,
     val element: XElement,
-    val annotatedClass: XTypeElement
+    val annotatedClass: XMemberContainer
```
rossbacher, line 27: This is the main fix.

BenSchwab (approved):
Ran into the same thing with scabbard

PR #328, opened 2021-08-16
Added support for allowed placeholder values

rossbacher:
Added support for allowed placeholder values by adding them viai `{placeholder(allowed|value)}` into the placeholder.

An URI that would otherwise match, but has a value for the placeholder that is not supported will not match.

This is mainly useful for scheme/host placeholders.


BenSchwab (approved):
Thanks @rossbacher - lgtm. 

My only food for though is if it would be better injected at runtime via DeepLinkDelegate, the same way configurablePathSegments are (the concept is actually pretty similar, right?). 

The benefit I see is in an app where the domains are declared in some other location, or dynamically, it may be easier to wire that up from with runtime code vs. needing some system to generate/update the DLD spec.

rossbacher:
> Thanks @rossbacher - lgtm.
> 
> My only food for thought is if it would be better injected at runtime via DeepLinkDelegate initialization, the same way `configurablePathSegments` are (the concept is actually pretty similar, right?).
> 
> The benefit I see is in an app where the domains are declared in some other location, or dynamically, it may be easier to wire that up from with runtime code vs. needing some system to generate/update the DLD spec.

@BenSchwab I like that with this inline definition they are in the same place. Let me see if it is easy to add that injections via the delegate in addition so one can choose which way of doing it.

PR #330, opened 2021-08-30
Added a new concept of deep link handlers.

rossbacher:
For the addition to the `README.md`:

You can annotate a Kotlin `object` that is extending `com.airbnb.deeplinkdispatch.handler.DeepLinkHandler`
with an `@DeepLink` annotation. e.g.DeepLinkHandler.

```kotlin
@DeepLink("foo://example.com/handlerDeepLink/{param1}?query1={queryParameter}")
object ProjectDeepLinkHandler : DeepLinkHandler<ProjectDeepLinkHandlerArgs>() {
    override fun handleDeepLink(parameters: ProjectDeepLinkHandlerArgs) {
        /**
         * From here any internal/3rd party navigation framework can be called the provided args.
         */
    }
}
data class ProjectDeepLinkHandlerArgs(
    @DeeplinkParam("param1", DeepLinkParamType.Path) val number: Int,
    @DeeplinkParam("query1", DeepLinkParamType.Query) val flag: Boolean?,
)
```

DeepLinkDispatch will then call `handleDeepLink` function in your handler with the path placeholders
and queryParameters converted into an instance of the specified type class.

This will give compile time safety as all placeholders and query parameters specified in the template
inside the `@DeepLink` annotation must be present in the arguments class for the processor to pass.
This is also true the other way around as all fields in the arguments class must be annotated and must
be present in the template inside the annotation.

From this function side the handler you can now call into any internal or 3rd party navigation system
without any Intent being fired at all and with type safety for your arguments.DeepLink

*Note:* That even though they must be listed in the template and annotation argument values annotated
with `DeepLinkParamType.Query` can be null as they are allowed to not be present in the matched url.
Type values that cannot be converted (e.g. a value that cannot be converted to a number) 0 is assumed
for non nullable `DeepLinkParamType.Path` values and null is returned for `DeepLinkParamType.Query` values.

## Changes

* Converted BaseDeepLinkDelegate to Kotlin which has some impact on Nullability.
* Converted `DeepLinkEntry` to a sealed class with 3 subtypes: `ActivityDeeplinkEntry`, `MethodDeeplinkEntry` and `HandlerDeeplinkEntry1
* Updated index to store the deep link entry type inside the index.
*


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
```diff
@@ -197,32 +238,180 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment?
         ) + (element.getAnnotation(DEEP_LINK_CLASS)?.value?.value?.toList() ?: emptyList())
     }
 
-    private fun verifyMethod(methodElement: XMethodElement) {
-        if (!methodElement.isStatic()) {
+    private fun verifyCass(classElement: XTypeElement): Boolean {
```
BenSchwab, line 241: It might be worth setting up a structure where these validations methods can throw, and a root try catch for a DLDException is caught, so we can avoid boolean return types used to early terminate a function

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
```diff
@@ -197,32 +238,180 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment?
...
-        }
+            false
+        } else
         // FIXME This is crashing with an NPE on internal classes when accessing the returnType.
         // You can jut comment this check out for now if you need this to pass.
-        if (methodElement.returnType.typeElement?.qualifiedName !in listOf(
-                "android.content.Intent",
-                "androidx.core.app.TaskStackBuilder",
-                "com.airbnb.deeplinkdispatch.DeepLinkMethodResult"
+            if (methodElement.returnType.typeElement?.qualifiedName !in listOf(
```
BenSchwab, line 272: Pull the list to the companion object to avoid re-allocation?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
```diff
@@ -197,32 +238,180 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment?
...
+                    "${com.airbnb.deeplinkdispatch.handler.DeepLinkHandler::class.qualifiedName}" +
+                    "' can be annotated with @DeepLink or another custom deep link annotation."
+            )
+            false
+        } else true
+    }
+
+    private fun validClassElement(classElement: XTypeElement) =
+        classElement.inheritanceHierarchyContains(
+            listOf(
```
BenSchwab, line 256: Pull the list to the companion object to avoid re-allocation?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
```diff
@@ -197,32 +238,180 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment?
...
+        val annotatedPathParameterNames = allPathParameters.mapNotNull {
+            it.getAnnotation(DeeplinkParam::class)?.value?.name
+        }.toSet()
+        if (annotatedPathParameterNames != templateHostPathSchemePlaceholders) {
+            logError(
+                element = argsType.typeElement ?: element,
+                message = "All scheme/host/path placeholders in the uri template must be annotated in the argument class constructor. " +
+                    "Present in urlTemplate: ${templateHostPathSchemePlaceholders.joinToString()} " +
+                    "Present in constructor: ${annotatedPathParameterNames.joinToString()}"
+            )
```
BenSchwab, line 350: Does this need a return here (or update `logError` to `throwError`)?

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
```diff
@@ -197,32 +238,180 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment?
...
         ) {
             logError(
-                element = methodElement,
-                message = (
-                    "Only `Intent`, `androidx.core.app.TaskStackBuilder` or " +
-                        "'com.airbnb.deeplinkdispatch.DeepLinkMethodResult' are supported. Please double " +
-                        "check your imports and try again."
-                    )
+                element = element,
+                message = "More than one method with a single parameter and $DEEP_LINK_HANDLER_METHOD_NAME name found in handler class."
```
BenSchwab, line 380: This should be impossible, right?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,442 @@
...
+     * Create a [DeepLinkResult], whether we are able to match the uri on
+     * {@param sourceIntent} or not.
+     *
+     * @param activity      used to startActivity() or notifyListener().
+     * @param sourceIntent  inbound Intent.
+     * @param deeplinkMatchResult deeplinkMatchResult that matches the sourceIntent's URI. Derived from
+     * [.findEntry]. Can be injected for testing.
+     * @return DeepLinkResult
+     */
+    fun createResult(
```
BenSchwab, line 203: Do we need to make dispatchResult public? Or an extension of `DeepLinkResult`? I'm guessing a use case may be to call `createResult` and the conditionally dispatch, and it would be nice to avoid going through the parsing logic a second time

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,442 @@
...
+    private fun dispatchResult(
+        result: DeepLinkResult,
+        activity: Activity
+    ) {
+        result.deepLinkMatchResult?.let { matchedDeepLinkResult ->
+            when (matchedDeepLinkResult.deeplinkEntry) {
+                is DeepLinkEntry.MethodDeeplinkEntry -> result.methodResult.taskStackBuilder?.startActivities()
+                    ?: activity.startActivity(result.methodResult.intent)
+                is DeepLinkEntry.ActivityDeeplinkEntry -> activity.startActivity(result.methodResult.intent)
+               is DeepLinkEntry.HandlerDeepLinkEntry -> if (result.isSuccessful) callDeeplinkHandler(result)
```
BenSchwab, line 109: Formatting looks a bit off

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,442 @@
...
+
+    private fun callDeeplinkHandler(result: DeepLinkResult) {
+        val handlerClazz = result.deepLinkMatchResult?.deeplinkEntry?.clazz!!
+        val handlerMethod = handlerClazz.methods.singleOrNull { method ->
+            method.name == DEEP_LINK_HANDLER_METHOD_NAME && method.genericParameterTypes.let { parameterTypes ->
+                parameterTypes.singleOrNull()?.let { it != Object::class.java } ?: false
+            }
+        } ?: error("DeepLinkHandler class ${handlerClazz.name} has more than one methods " +
+                "\"$DEEP_LINK_HANDLER_METHOD_NAME\" methods with a single parameter.")
+        // Ok to call single here as we would have failed above if we would have more than one type.
```
BenSchwab, line 122: Could we cast the class to `Class<out DeepLinkHandler>` to avoid the need to very the number of methods?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,442 @@
...
+        val handlerParameterClazz = handlerMethod?.genericParameterTypes?.single() as Class<*>
+        val handlerParameterClazzConstructor = handlerParameterClazz.constructors.singleOrNull()
+            ?: error("Handler parameter class can only have one constructor.")
```
BenSchwab, line 125: lazy q: Is this verified at annotation processing time as well?

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,442 @@
...
+            // Check if we have as many parameters as annotations, the DeeplinkParam annotation is not
+            // repeatable we cannot have multiple for just one element.
+            if (annotationList.size == it.size) {
+                annotationList.zip(it)
+            } else error("There are ${annotationList.size} annotations but ${it.size} parameters!")
+        }
+        val handlerArgsConstructorParams = createParamArray(typeNameMap, result.parameters)
+        val handlerParameters =
+            handlerParameterClazzConstructor.newInstance(*handlerArgsConstructorParams)
+        val handlerInstanceField = handlerClazz.getField("INSTANCE").get(null);
```
BenSchwab, line 140: Does this mean we only support `objects`? I think we could pretty easily also support a 0-arg constructor (which would provide better backwards support for java)

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,442 @@
...
+    private fun callDeeplinkHandler(result: DeepLinkResult) {
+        val handlerClazz = result.deepLinkMatchResult?.deeplinkEntry?.clazz!!
+        val handlerMethod = handlerClazz.methods.singleOrNull { method ->
+            method.name == DEEP_LINK_HANDLER_METHOD_NAME && method.genericParameterTypes.let { parameterTypes ->
+                parameterTypes.singleOrNull()?.let { it != Object::class.java } ?: false
+            }
+        } ?: error("DeepLinkHandler class ${handlerClazz.name} has more than one methods " +
+                "\"$DEEP_LINK_HANDLER_METHOD_NAME\" methods with a single parameter.")
+        // Ok to call single here as we would have failed above if we would have more than one type.
+        val handlerParameterClazz = handlerMethod?.genericParameterTypes?.single() as Class<*>
```
BenSchwab, line 123: I don't think handlerMethod can be null here -- (as an aside, maybe we should configure kotlin errors as warnings for this repo?)

BenSchwab (commented):
This is really exciting - love that it can be adapted to any navigation framework.

Did you consider generating the arg parsing functions? the reflective code looks fairly complex, and it might be easier to just to generate an `[ArgsClass]Parser`, which would be more performant, and offer a compiler checks for some of the edge cases around nullability

BenSchwab (approved):
Discussed some of the questions I have offline.

Where I see we landed:
- probably not worth generating parsing classes
- We can probably unify query/param annotations, but having an explicit key value provided is good for consistency (as it's required for query values)
- Can implement some logic to hydrate URLs given an argument class an a handler
- We can follow up to allow matching directly to a combination of a parsed args class + handler class to allow even more flexible routing, such as showing a fragment

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
rossbacher: @BenSchwab No, see the unit test testing that(1testDeeplinkHandlerWithTwoSingleParameterMethodsWithCorrectName`). It can have a different parameter type. And as the generics get deleted during runtime I don't know which of them would be the correct type if there are multiple.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: @BenSchwab yes:

```kotlin
val argsConstructor = argsType.typeElement?.getConstructors()?.singleOrNull() ?: run {
    logError(
        element = argsType.typeElement ?: element,
        message = "Argument class can only have one constructor"
    )
    return
}
```

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: Added support for zero argument constrcutors and added tests for it as well.
There were also some issues with Java code in general and I added tests that test Java code, plus fixes to make it work.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: We would still need to check the methods as there can be multiple and I do not have access to the generic type here.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: I'll do that as a separate PR as discussed offline.

README.md
```diff
@@ -48,6 +48,43 @@ public class MainActivity extends Activity {
...
+inside the `@DeepLink` annotation must be present in the arguments class for the processor to pass.
+This is also true the other way around as all fields in the arguments class must be annotated and must
+be present in the template inside the annotation.
+
+From this function side the handler you can now call into any internal or 3rd party navigation system
+without any Intent being fired at all and with type safety for your arguments.DeepLink
+
+*Note:* That even though they must be listed in the template and annotation argument values annotated
+with `DeepLinkParamType.Query` can be null as they are allowed to not be present in the matched url.
+Type values that cannot be converted (e.g. a value that cannot be converted to a number) 0 is assumed
```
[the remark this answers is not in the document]
rossbacher: @BenSchwab I added support for configuring this.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
BenSchwab: Don't we verify in the annotation processor that there is not multiple methods of that name?

I think casting to the class should be possible  `Class<out DeepLinkHandler<Any>>`, and would avoid needing a keep rule for the handle method name

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -0,0 +1,465 @@
...
+        val typeNameMap = handlerParameterClazzConstructor.parameterTypes.let {
+            // Check if we have as many parameters as annotations, the DeeplinkParam annotation is not
+            // repeatable we cannot have multiple for just one element.
+            if (annotationList.size == it.size) {
+                annotationList.zip(it)
+            } else error("There are ${annotationList.size} annotations but ${it.size} parameters!")
+        }
+
+        val handlerInstance = try {
+            handlerClazz.getField("INSTANCE").get(null)
```
BenSchwab, line 142: Do we need to update any project keep rules?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/handler/Annotations.kt
```diff
@@ -0,0 +1,8 @@
+package com.airbnb.deeplinkdispatch.handler
+
+@Target(AnnotationTarget.VALUE_PARAMETER, )
+@Retention(AnnotationRetention.RUNTIME)
+@MustBeDocumented
+annotation class DeeplinkParam(val name: String, val type: DeepLinkParamType)
```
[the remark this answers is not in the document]
rossbacher: Yes, I think it is just more explicit this way.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: We do need to update them, good catch. I thought about it but then forgot.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: Problem is that I need the method to get the type of the arg object. I don't think there is a way around the keep rule. I still did the casting to make the calling simpler and not using reflection for it.

PR #335, opened 2021-10-20
Fix issues with long placeholders in hosts.

rossbacher:
Hosts can be very long if they have placeholders (in templates) and this might not process correctly. So we replace the placeholders with short ones and rereplace them after processing.

This is safe as this is only done for templates and placeholder format is well understood.

Added tests

### DIFF
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
@@ -1217,8 +1217,8 @@ public final class DeepLinkUri {
      */
     private static String domainToAscii(String input) {
       try {
-        // When this is a template URI it might get to long with all the placeholder definitions
-        // to pass verification.
+        // When this is a template URI it might get to long (> 256 chars) with all the placeholder
+        // definitions to pass verification.
         // This will replace the placeholders with other shorter placeholders for URI validation.
         Set<String> placeholders = getPlaceHolders(input);
         String mappedInput = input;
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
@@ -1221,13 +1221,13 @@ public final class DeepLinkUri {
         String mappedInput = input;
         Map<String, String> placeholderMap = new HashMap(placeholders.size());
         int i = 0;
-        for(String placeholder: placeholders) {
-          String replacedValue = ""+(i++);
+        for (String placeholder : placeholders) {
+          String replacedValue = "" + (i++);
           mappedInput = mappedInput.replace(placeholder, replacedValue);
           placeholderMap.put(placeholder, replacedValue);
         }
         String result = IDN.toASCII(mappedInput).toLowerCase(Locale.US);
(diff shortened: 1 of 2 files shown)


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
```diff
@@ -1215,7 +1217,22 @@ private static boolean decodeIpv4Suffix(
...
       try {
-        String result = IDN.toASCII(input).toLowerCase(Locale.US);
+        // When this is a template URI it might get to long with all the placeholder definitions
+        // to pass verification.
+        // This will replace the placeholders with other shorter placeholders for URI validation.
+        Set<String> placeholders = getPlaceHolders(input);
+        String mappedInput = input;
+        Map<String, String> placeholderMap = new HashMap(placeholders.size());
+        int i = 0;
+        for (String placeholder : placeholders) {
```
elihart, line 1227: would be really nice to have this file in kotlin (not in this PR of course, but one day :P )

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
```diff
@@ -1215,7 +1217,22 @@ private static boolean decodeIpv4Suffix(
      */
     private static String domainToAscii(String input) {
       try {
-        String result = IDN.toASCII(input).toLowerCase(Locale.US);
+        // When this is a template URI it might get to long with all the placeholder definitions
```
elihart, line 1220: do you have a number for what the length limit is?

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
```diff
@@ -1215,7 +1217,22 @@ private static boolean decodeIpv4Suffix(
...
+        // to pass verification.
+        // This will replace the placeholders with other shorter placeholders for URI validation.
+        Set<String> placeholders = getPlaceHolders(input);
```
elihart, line 1223: this could probably be pulled out to a standalone function that could encapsulate its functionality and be more directly testable

elihart (approved):

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
rossbacher: Well the reason it is not is because it is literally a copy of another file. That is why I never ported it.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
rossbacher: 256. I'll add it to the comment.

deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkUri.java
rossbacher: In Kotlin yes, in Java it is a bit tricky as there would be multiple returns if I don't create a wrapper. I'll leave this like this for now but if we ever convert this to Kotlin it makes sense.

PR #336, opened 2021-11-01
Bump versions to fix xprocessing API change

elihart:
The new version of room/xprocessing has a breaking api change that is not backwards compatible (making a nullable param non null). Since we have updated to this new version in our other libraries it pushes the dependency up for this library as well, causing a compilation error. to solve we need to update in all libraries.

I also took some time to update other libraries too, notably AGP.

Tests also started failing, likely because of the xprocessing update, because I think the "overrides" function was using the wrong owner param. Not sure why it worked before - it's possible xprocessing changed this api behavior too.

### DIFF
--- a/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
+++ b/deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
@@ -298,9 +298,10 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment? 
         val argsType = element.getAllMethods().singleOrNull {
             it.overrides(
                 other = handleDeepLinkInterfaceMethod,
-                owner = deepLinkHandlerInterface
+                owner = element
             )
-        }?.parameters?.last()?.type ?: error("Is not overriding method from interface. This is impossible.")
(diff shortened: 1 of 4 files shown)


deeplinkdispatch-processor/build.gradle
```diff
@@ -26,7 +25,6 @@ dependencies {
   testImplementation deps.compileTestingKsp
   testImplementation deps.mockk
   testImplementation deps.kotlinReflect
-  testImplementation files("${System.properties['java.home']}/../lib/tools.jar")
```
rossbacher: How did you work around this? This line was bugging me for so long but I had no idea how to loose that dependency.

deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkProcessor.kt
```diff
@@ -298,9 +298,10 @@ class DeepLinkProcessor(symbolProcessorEnvironment: SymbolProcessorEnvironment?
         val argsType = element.getAllMethods().singleOrNull {
             it.overrides(
                 other = handleDeepLinkInterfaceMethod,
-                owner = deepLinkHandlerInterface
```
rossbacher, line 301: Hm, also no idea how this worked. Thx for the fixes!

rossbacher (approved):

PR #341, opened 2021-11-17
Ensure that entries in allowed placeholder values are sorted

rossbacher:
Make sure that allowed placeholder values inside of index are sorted, as we using binary search to look them up.

Add tests for both the function that does the sorting and also for the index itself to see if values get correctly sorted into the index.


elihart (approved):
Thanks!

PR #342, opened 2021-11-18
Fix crash that could happen with no arg deep Link handler in R8 full mode.

rossbacher:
The bytecode in the `.dex` file for the DeepLinkHander interface looks like this after normal compilation:

```smali
###### Class com.airbnb.deeplinkdispatch.handler.DeepLinkHandler (com.airbnb.deeplinkdispatch.handler.DeepLinkHandler)
.class public interface abstract Lcom/airbnb/deeplinkdispatch/handler/DeepLinkHandler;
.super Ljava/lang/Object;
.source "DeepLinkHandler.kt"


# annotations
.annotation system Ldalvik/annotation/Signature;
    value = {
        "<T:",
        "Ljava/lang/Object;",
        ">",
        "Ljava/lang/Object;"
    }
.end annotation
```

If we run R8 with full mode enabled on this it looks like this:

```smali
###### Class com.airbnb.deeplinkdispatch.handler.DeepLinkHandler (com.airbnb.deeplinkdispatch.handler.DeepLinkHandler)
.class public interface abstract Lcom/airbnb/deeplinkdispatch/handler/DeepLinkHandler;
.super Ljava/lang/Object;
```

The `dalvik/annotation/Signature` that describes the generic type got removed. (See https://r8.googlesource.com/r8/+/refs/heads/main/compatibility-faq.md).

This is no issue at runtime as this is not needed and the generic type of anything implementing the interface is still kept.

e.g.

```smali
###### Class com.airbnb.deeplinkdispatch.sample.handler.SampleKotlinDeepLinkHandler (com.airbnb.deeplinkdispatch.sample.handler.SampleKotlinDeepLinkHandler)
.class public final Lcom/airbnb/deeplinkdispatch/sample/handler/SampleKotlinDeepLinkHandler;
.super Lcom/airbnb/deeplinkdispatch/sample/handler/IntermediateDeepLinkHandler1;


# annotations
.annotation runtime Lcom/airbnb/deeplinkdispatch/sample/WebDeepLink;
.end annotation

.annotation system Ldalvik/annotation/Signature;
    value = {
        "Lcom/airbnb/deeplinkdispatch/sample/handler/IntermediateDeepLinkHandler1<",
        "Lcom/airbnb/deeplinkdispatch/sample/handler/TestKotlinDeepLinkHandlerDeepLinkArgs;",
        ">;"
    }
.end annotation
```

but the type is fully erased if it is just `Any` e.g.

```smali
###### Class com.airbnb.deeplinkdispatch.sample.handler.SampleNoParamsKotlinDeepLinkHandler (com.airbnb.deeplinkdispatch.sample.handler.SampleNoParamsKotlinDeepLinkHandler)
.class public final Lcom/airbnb/deeplinkdispatch/sample/handler/SampleNoParamsKotlinDeepLinkHandler;
.super Ljava/lang/Object;

# interfaces
.implements Lcom/airbnb/deeplinkdispatch/handler/DeepLinkHandler;


# annotations
.annotation runtime Lcom/airbnb/deeplinkdispatch/sample/WebDeepLink;
.end annotation

.annotation system Ldalvik/annotation/Signature;
    value = {
        "Ljava/lang/Object;",
        "Lcom/airbnb/deeplinkdispatch/handler/DeepLinkHandler;"
    }
.end annotation

```

Directly keeping the `DeepLinkHandler` class is fixing this issue.
I also added some other code to make this safer and allow error reporting when type determination does not work.

- Keep deep link handler so tje signature attributes are also kept in the `.dex` file.
- Make `ErrorHandler` an abstract class so we can add error when type of deeplink handler args class cannot be determined.
- Fall back to `Any::class.java` when no type can be determined.
- Switch on R8 full mode to allow visibility into errors caused by full mode.
- Update gradle to latest.

### DIFF
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/ErrorHandler.kt
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/ErrorHandler.kt
@@ -1,5 +1,13 @@
 package com.airbnb.deeplinkdispatch
 
-interface ErrorHandler {
-    fun duplicateMatch(uriString: String, duplicatedMatches: List<DeepLinkMatchResult>)
+abstract class ErrorHandler {
+    open fun duplicateMatch(uriString: String, duplicatedMatches: List<DeepLinkMatchResult>) {}
+
+    /**
+     * Called if unable to determine the type of an args class for a DeepLinkHandler.
+     * If this is the case we are just assuming no arguments and do not fail this might
+     * but this should not happen.
+     * This will give youyr pooportunity to log these cases.
+     */
+    open fun unableToDetermineHandlerArgsType(uriTemplate: String, className: String) {}
 }
--- a/deeplinkdispatch/proguard-rules.pro
+++ b/deeplinkdispatch/proguard-rules.pro
@@ -1,8 +1,16 @@
 # Keep the deep link handler classes (and INSTANCE field if they have it) as they are accessed
 # via reflection from within the app.
--keep class * extends com.airbnb.deeplinkdispatch.handler.DeepLinkHandler {
+-keep class * implements com.airbnb.deeplinkdispatch.handler.DeepLinkHandler {
     public static final ** INSTANCE;
 }
+
+# As per https://r8.googlesource.com/r8/+/refs/heads/main/compatibility-faq.md in full mode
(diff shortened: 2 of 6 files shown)


BenSchwab (approved):
Thanks for digging into this!

PR #343, opened 2021-11-29
Address xprocessing api change

elihart:
There was this change in xprocessing to wrap annotation list values in JavacAnnotationValue

before:
```
    override fun visitArray(vals: MutableList<out AnnotationValue>, env: JavacProcessingEnv) =
        vals.map { it.accept(this, env) }
```

after:

```
    override fun visitArray(vals: MutableList<out AnnotationValue>, data: VisitorData) =
        vals.map { JavacAnnotationValue(data.env, data.method, it) { it.accept(this, data) } }
```

To make this safe so that it works with multiple versions of xprocessing depending on where the library is used (and since the breaking change is only in the xprocessing snapshot version right now) I have updated DLD to work in both cases


deeplinkdispatch-processor/src/main/java/com/airbnb/deeplinkdispatch/ProcessorUtils.kt
```diff
@@ -50,3 +53,19 @@ fun XTypeElement.directlyImplementsInterfaces(fqnList: List<String>): Boolean {
         getSuperInterfaceElements().any { typeElement -> typeElement.qualifiedName == interfaceFqn }
     }
 }
+
+inline fun <reified T> XAnnotation.getAsList(method: String): List<T> {
+    val originalList = get<List<T>>(method)
+    // In new XProcessing versions List values are wrapped in XAnnotationValue but in old versions
```
BenSchwab, line 59: Maybe worth putting a todo in the xprocessing version that when it is bumped that this can be simplified.

BenSchwab (approved):

gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=6.0.0-beta07-SNAPSHOT
```
rossbacher, line 1: I guess this would still be beta 07 snapshot, right as we did not release the 07 version yet

rossbacher (approved):

PR #347, opened 2022-03-08
Update dependencies / Clarify parameter override behavior

rossbacher:
Ensure that bundle extra values set in the deeplink method always override values that are set via the URL or via Intent extras.

Added tests to clarify this behavior this.

Update everything except robolectric to latest.

### DIFF
--- a/deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
+++ b/deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
@@ -304,7 +304,7 @@ open class BaseDeepLinkDelegate @JvmOverloads constructor(
                 if (intent.data == null) {
                     intent.data = sourceIntent.data
                 }
-                intent.putExtras(intentBundle.filter { key, _ -> intent.extras?.let { !it.containsKey(key) } ?: true })
+                intent.putExtras(intentBundle.filter { key, _ -> intent.extras?.containsKey(key) != true })
                 intent.putExtra(DeepLink.IS_DEEP_LINK, true)
                 intent.putExtra(DeepLink.REFERRER_URI, originalIntentUri)
                 if (activity.callingActivity != null) {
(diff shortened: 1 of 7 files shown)


deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -304,7 +304,7 @@ open class BaseDeepLinkDelegate @JvmOverloads constructor(
                 if (intent.data == null) {
                     intent.data = sourceIntent.data
                 }
-                intent.putExtras(intentBundle)
+                intent.putExtras(intentBundle.filter { key, _ -> intent.extras?.let { !it.containsKey(key) } ?: true })
```
elihart, line 307: I think this is what you want?
```suggestion
                intent.putExtras(intentBundle.filter { key, _ -> intent.extras?.containsKey(key) != true )
```

dependencies.gradle
```diff
@@ -1,20 +1,21 @@
 def versions = [
-    kotlinVersion                : '1.5.31',
-    appCompatVersion             : '1.2.0',
-    localBroadcastManagerVersion : '1.0.0',
+    kotlinVersion                : '1.6.10',
```
elihart, line 2: thanks for all the updates too

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/Utils.kt
```diff
@@ -23,3 +25,17 @@ fun validateConfigurablePathSegmentReplacements(
             "are:\n${configurablePathSegmentReplacements.keys.joinToString(",\n") { String(it) }}."
     }
 }
+
+/**
+ * Returns a shallow copy of this Bundle that is missing the items for which the predicate returned
+ * false
+ */
+fun Bundle.filter(predicate: (key: String, value: Any?) -> Boolean): Bundle {
```
elihart, line 33: unit test for this specifically?

elihart (approved):
thanks!

mmdango (approved):
Thanks!

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
rossbacher: Yeah, I did have it kinda like that yesterday (before reverting to the ugly way of doing it but I could just not wrap my mind around it. Thank you!

Issue #351, opened 2022-07-04
queryParameterValue must not be null


masonHong:
I was using version **5.4.3**, but when I upgraded to version **6.1.0**, a crash occurred, so I checked it.

```
Fatal Exception: java.lang.NullPointerException: queryParameterValue must not be null
       at com.airbnb.deeplinkdispatch.BaseDeepLinkDelegate.queryAndPathParameters(BaseDeepLinkDelegate.java:530)
       at com.airbnb.deeplinkdispatch.BaseDeepLinkDelegate.createResult(BaseDeepLinkDelegate.java:286)
       at com.airbnb.deeplinkdispatch.BaseDeepLinkDelegate.dispatchFrom(BaseDeepLinkDelegate.java:70)
       at com.airbnb.deeplinkdispatch.BaseDeepLinkDelegate.dispatchFrom$default(BaseDeepLinkDelegate.java:64)
```

When I reproduced it, a crash occurs when `Uri` of the pattern below is used.

- myScheme://host/path1/path2?query1&query2
- myScheme://host/path1/path2?

Uri have a query with no value or with `?` but no query.

Can you fix the problem?

jackcsk:
@EricKuck are we expecting a new release including that fix soon?

EricKuck:
I am not a maintainer on this project. Can't say if/when a new release is coming out.

jackcsk:
My bad, sorry Eric.

jackcsk:
@rossbacher : not sure if you are the right person to address to, or please help me to refer to the right person
Wanna know when this fix will be included in the new release

masonHong:
This issue fixed on 5ed4682. version 6.2.2

PR #352, opened 2022-07-19
Fix NPE in issue #351

EricKuck:
Issue #351 reports an NPE when query parameters do not have values, such as `schema://mycompany/search?q`. This PR fixes that issue by using empty strings for this case.


rossbacher:
Really appreciate the fix did not get to that issue yet but could you also add a unit test so that issue does not slip through the cracks again?

EricKuck:
Sure, test added

rossbacher (approved):

rossbacher:
Thank you @EricKuck!

PR #355, opened 2022-10-27
Support for easily finding duplicates

rossbacher:
Add support for easily finding duplicate definitions across modules via a unit test in the app.

### DIFF
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
@@ -33,35 +33,9 @@ data class DeepLinkMatchResult(
 
     override fun toString(): String {
         return "uriTemplate: ${deeplinkEntry.uriTemplate} " +
-            "activity: ${deeplinkEntry.clazz.name} " +
-            "${if (deeplinkEntry is DeepLinkEntry.MethodDeeplinkEntry) "method: ${deeplinkEntry.method} " else ""}" +
-            "parameters: $parameterMap"
-    }
-
-    private val firstConfigurablePathSegmentIndex: Int by lazy {
-        deeplinkEntry.uriTemplate.indexOf(
-            configurablePathSegmentPrefixChar
-        )
-    }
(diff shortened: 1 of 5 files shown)


deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/DeepLinkEntry.kt
```diff
@@ -38,13 +38,64 @@ data class DeepLinkMatchResult(
...
     }
 
+    /**
+     * Whatever template has the first placeholder (and then configurable path segment) is the less
+     * concrete one.
+     * Because if they would have been all in the same index those elements would have been on the
+     * same level and in the same "list" of elements we compare in order.
+     * In this case the one with the more concete element would have won and the same is true here.
+     */
+    override fun compareTo(other: DeepLinkMatchResult): Int {
```
rossbacher, line 48: I just moved the compare logic from `DeepLinkMatchResult` to `DeepLinkEntry` as I wanted to use that first but then ended up not doing it, however the change still makes sense as the compare is just based on the `DeepLinkEntry` anyway.

deeplinkdispatch/src/main/java/com/airbnb/deeplinkdispatch/BaseDeepLinkDelegate.kt
```diff
@@ -583,6 +583,21 @@ open class BaseDeepLinkDelegate @JvmOverloads constructor(
         registries.flatMap { it.getAllEntries() }
     }
 
+    /**
+     * Get a map of all DeepLinkEntries and its duplicates, DeepLinkEntry objects that
+     * might be slightly different but will map to the same URL during app operation.
+     */
+    val duplicatedDeepLinkEntries: Map<DeepLinkEntry, List<DeepLinkEntry>> by lazy {
+        allDeepLinkEntries.mapNotNull { deepLinkEntry ->
```
elihart, line 591: it may be nice to extract this function to be top level and take a list of deeplink entry and return the duplicates map.

In our test we have the set of registries, not this delegate

elihart (approved):
thanks! one suggestion

PR #357, opened 2022-11-15
Update kotlin and ksp

elihart:
Also fixes an issue with latest version of xprocessing that broke how handler override lookup was done.

Finally, updates to the latest publish plugin and improves the pattern we use for publishing.


BenSchwab (approved):

rossbacher (approved):
Thank you!

PR #364, opened 2023-04-05
Fix snapshot release

rossbacher:
Append secrets to existing `gradle.properites` file to avoid having to print them on cli (which would be redacted but still)

Update snapshot publish

### DIFF
--- a/.github/workflows/deploy_snapshot.yml
+++ b/.github/workflows/deploy_snapshot.yml
@@ -16,12 +16,12 @@ jobs:
         uses: actions/checkout@v2
       - name: Update gradle.properties
         uses: DamianReeves/write-file-action@master
-          with:
-            path: gradle.properties
-            write-mode: append
-            contents: |
-              mavenCentralUsername=${{ secrets.NEXUS_USER }}
-              mavenCentralPassword=${{ secrets.NEXUS_PASSWORD }}
+        with:
+          path: gradle.properties
+          write-mode: append
+          contents: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_snapshot.yml
```diff
@@ -16,7 +25,4 @@ jobs:
         uses: actions/checkout@v2
       - name: Deploy new snapshot of DeepLinkDispatch
         shell: bash
-        env:
-          SONATYPE_NEXUS_USERNAME: ${{ secrets.NEXUS_USER }}
-          SONATYPE_NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
-        run: ./gradlew uploadArchives
+        run: ./gradlew publishAllPublicationsToMavenCentral
```
elihart, line 28: if someone changes the version to a non snapshot version number, then this will publish that real version (since it has the ability to close and release the repository, which previously didn't exist)

Maybe we also need to verify the version ends in -SNAPSHOT before running this action?

docs https://vanniktech.github.io/gradle-maven-publish-plugin/central/#publishing-snapshots

publishing.gradle
```diff
@@ -20,6 +20,11 @@ publishing {
 }
 
 mavenPublishing {
+    if (findProperty("enforceSnapshotVersion").toString().toBoolean() && !findProperty("VERSION_NAME").toString().endsWith("-SNAPSHOT")) {
```
elihart, line 23: thanks!

elihart (approved):

Issue #350, opened 2022-06-24
Performance issue in chunkOnModifiedUtf8ByteSize


rciovati:
While debugging some built time issues in our codebase I noticed that `DeepLinkProcessor` takes a suspiciously high amount of time (in our case 3x than Dagger). I grabbed a trace with YourKit and, if I'm reading it correctly, the processor is spending most of its time in `chunkOnModifiedUtf8ByteSize`.

![Screen Shot 2022-06-24 at 13 38 19](https://user-images.githubusercontent.com/245528/175527765-a21df4f0-f5b4-48ab-9b44-61957dfb070e.png)

![Screen Shot 2022-06-24 at 13 45 03](https://user-images.githubusercontent.com/245528/175527947-688012ce-55b3-4057-8c94-d45b1aefa5d5.png)

We're not using ksp yet but it seems to me that this code path is shared between the apt and ksp so I suspect the issue exists there too.

rossbacher:
fixed with #386

PR #386, opened 2026-01-23
Optimize chunkOnModifiedUtf8ByteSize from O(n²) to O(n)

Nepomuk5665:
## Summary

- Fixes O(n²) performance issue in `chunkOnModifiedUtf8ByteSize` that caused significant build time overhead
- Implements incremental byte size calculation instead of creating substrings for each character
- Adds a helper function to calculate Modified UTF-8 byte size per character

## Background

This addresses the performance issue reported in #350. The original implementation created a new substring and converted it to a byte array for **every character** in the input string:

```kotlin
// Before: O(n²) - substring + toByteArray for each iteration
val charModifiedUtf8ByteArraySize =
    this
        .substring(nextChunkStart, i + 1)
        .let { chunk -> chunk.toByteArray().size + chunk.count { char -> char == '\u0000' } }
```

This is now replaced with an O(n) approach that calculates byte sizes incrementally:

```kotlin
// After: O(n) - direct calculation per character
val charByteSize = this[i].modifiedUtf8ByteSize()
```

## Technical Details

The new `modifiedUtf8ByteSize()` helper function handles Modified UTF-8 encoding rules:
- U+0000 (null): 2 bytes (C0 80 encoding - special for Modified UTF-8)
- U+0001-U+007F: 1 byte
- U+0080-U+07FF: 2 bytes
- U+0800-U+FFFF: 3 bytes

## Testing

All existing tests in `CharSequenceUtf8Test` pass, confirming behavioral correctness.

### DIFF
--- a/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
+++ b/deeplinkdispatch-base/src/main/java/com/airbnb/deeplinkdispatch/base/Utils.kt
@@ -17,7 +17,7 @@ import java.io.InputStream
 private fun Char.modifiedUtf8ByteSize(): Int {
     val codePoint = this.code
     return when {
-        codePoint == 0 -> 2  // Null character is encoded as 2 bytes in Modified UTF-8
+        codePoint == 0 -> 2 // Null character is encoded as 2 bytes in Modified UTF-8
         codePoint <= 0x7F -> 1
         codePoint <= 0x7FF -> 2
         else -> 3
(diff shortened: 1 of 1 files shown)


rossbacher:
This is great! thank you very much!

Can you fix the Ci issues? Just run `./gradlew check javaDoc` and fix the lint issues by running `formatKotlin` on the broken module,.

rossbacher (approved):
Thank you! Great fix!

PR #388, opened 2026-01-23
Centralize toolchain and target compatibility version.

rossbacher:
Use 22 for toolchain and 11 for source compatibility.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -12,9 +12,9 @@ jobs:
       - name: Checkout DeepLinkDispatch
         uses: actions/checkout@v2
       - name: Set up JDK 22
-        uses: [actions/setup-java](https://github.com/actions/setup-java)@v4
+        uses: actions/setup-java@v5
         with:
-          java-version: '22'
           distribution: 'coretto'
+          java-version: '22'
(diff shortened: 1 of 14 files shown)


.github/workflows/build_test.yml
```diff
@@ -11,5 +11,10 @@ jobs:
     steps:
       - name: Checkout DeepLinkDispatch
         uses: actions/checkout@v2
+      - name: Set up JDK 22
```
rossbacher, line 14: 25 did not work with robolectric and it was not worth to figure it out.

build.gradle
```diff
@@ -44,6 +44,53 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     }
 }
 
+// Apply consistent JVM toolchain and target to all subprojects
+subprojects {
```
elihart, line 48: kind of crazy how hard it is just to set a standard version for the whole project 😂

elihart (approved):
Great clean up, thanks 👍

build.gradle
rossbacher: Yep, but better to have it spread out everywhere...

PR #391, opened 2026-01-28
Add asset-based match index support for KSP library modules

rossbacher:
This change adds an optional asset-based approach for storing the match index binary data when using KSP with the manifest-generation plugin. Instead of encoding the binary data as strings in the generated registry class, the match index is written as an Android asset file and loaded at runtime via AssetManager.

Key changes:
- Add shared constants in ManifestGeneration.kt for asset paths and KSP option
- Generate AssetManager-based registry constructor when useAssetBasedMatchIndex=true
- Add MergeDeepLinkAssetsTask and RelocateDeepLinkAssetsTask for Gradle plugin
- Update ManifestGenerationPlugin to handle asset file relocation and merging
- Add processor tests for asset-based and legacy string-based code generation
- Update sample tests to use KspLibraryDeepLinkModuleRegistry with assets

The asset-based approach provides:
- Faster build times (no string chunking/encoding)
- Faster app startup (direct binary loading, no string decoding)
- Better APK compression (binary assets compress better than dex strings)

Both modes (asset-based and string-based) are supported per-registry level.

I did run benchmarks on this and in the registry creation (the important benchmark) the asset based approach is bout 20-25% faster, less than I had hoped but this also simplifies this significantly especially once we drop kapt support.

## Validation

- [GitHub Actions Build/Test](https://github.com/airbnb/DeepLinkDispatch/actions/runs/35786378820) passed for commit `ba836b0`.

- The complete CI task set (`assemble check javadoc lintKotlin`) passes locally with an empty Maven-local repository and a valid Android SDK: 544 test executions passed, with no failures or skips.
- Kotlin formatting and the generated-asset dependency for `generateDebugLintReportModel` are fixed.
- Android lint remains enabled. The Activity-inheritance errors encountered locally came from malformed SDK package metadata; validation used an isolated SDK with corrected metadata.


gradlew
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 #
-# Copyright © 2015 the original authors.
+# Copyright © 2015-2021 the original authors.
```
rossbacher, line 4: Just updated gradlew to latest as of the used gradle version (9.2.1)

elihart (approved):

PR #394, opened 2026-04-20
Manifest generation: fail on write errors, clean up logging

rossbacher:
## Summary
- ManifestGenerator now throws DeepLinkProcessorException on IOException when writing the generated AndroidManifest.xml, so write failures fail the build cleanly instead of being swallowed as a diagnostic warning that surfaces later in a confusing place. Catch is scoped to IOException so genuine processor bugs (NPE/ISE) are no longer hidden.
- RelocateDeepLinkManifestTask: use orNull on the @Optional kspManifestFile property (get() NPEs when unset), replace println with logger.warn so the message respects Gradle log level, collapse an empty else-if branch, and fix a couple of typos.
- GenerateManifestIntentFiltersForDeeplinkDispatchTask: replace println with logger.warn, and replace error(...) (IllegalStateException) with throw GradleException(...) for idiomatic build-failure output.

## Test plan
- [x] ./gradlew :deeplinkdispatch-processor:test :deeplinkdispatch-gradle-plugin:test
- [x] ./gradlew :sample:testDebugUnitTest :deeplinkdispatch:test :deeplinkdispatch-base:test

### DIFF
--- a/deeplinkdispatch-gradle-plugin/src/main/java/com/airbnb/deeplinkdispatch/gradleplugin/RelocateDeepLinkManifestTask.kt
+++ b/deeplinkdispatch-gradle-plugin/src/main/java/com/airbnb/deeplinkdispatch/gradleplugin/RelocateDeepLinkManifestTask.kt
@@ -41,20 +41,18 @@ abstract class RelocateDeepLinkManifestTask : DefaultTask() {
         // 2. Gradle's up-to-date check only compares content, not file existence
         // 3. The task deletes the source file, which needs to happen every time
         outputs.upToDateWhen {
-            // Task is only up-to-date if source file doesn't exist
-            !kspManifestFile.get().asFile.exists()
+            val source = kspManifestFile.orNull?.asFile
+            source == null || !source.exists()
         }
     }
 
     @TaskAction
     fun taskAction() {
-        val sourceFile = kspManifestFile.get().asFile
+        val sourceFile = kspManifestFile.orNull?.asFile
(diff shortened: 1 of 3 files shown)


elihart (approved):

deeplinkdispatch-gradle-plugin/src/main/java/com/airbnb/deeplinkdispatch/gradleplugin/RelocateDeepLinkManifestTask.kt
```diff
@@ -70,14 +68,13 @@ abstract class RelocateDeepLinkManifestTask : DefaultTask() {
...
-            // Source doesn't exist but dest does - this can happen on incremental builds
-            // where KSP was UP-TO-DATE and we already moved the file previously.
-            // The dest file is still valid, so nothing to do.
-        } else {
-            // Neither file exists - no manifest was generated
-            println("No DeepLinkDispatch manifest found to relocate in ${project.name}. If this module has no deep links, consider" +
-                    "removing the DeepLinkDispatch gradle plugin from it's gradle file.")
+        } else if (!destFile.exists()) {
+            logger.warn(
+                "No DeepLinkDispatch manifest found to relocate in ${project.name}. " +
```
elihart: what about the change to switch to path?

https://github.com/airbnb/DeepLinkDispatch/pull/396/changes#diff-d9f2798809f7b9170b8199da517f11caa1a2d58e2c1284521137da518e61d262L79

